PR #15, opened 2022-01-03
chore: Improve Code Style

yezz123:
@aminalaee I like the Idea behind SQLAdmin, I will add some features to it in the future and I will use it in one of my projects soon, While I read the code I found some parts that could be refactored and enhanced to be better.

If you have any questions, we could discuss them!

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,4 @@
-comment: off
+coverage:
+  precision: 1
+
+comment: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,8 @@
 coverage:
-  precision: 1
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
 
 comment: false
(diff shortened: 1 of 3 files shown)


aminalaee:
Hey @yezz123 ,
Glad you liked it and will use it.

Any feedback and contribution would be appreciated :)

aminalaee:
I think the codecov precision needs to be configured. I can do it in a separate PR, but if you're interested we can fix it here.

yezz123:
> I think the codecov precision needs to be configured. I can do it in a separate PR, but if you're interested we can fix it here.

Yes That's would be great!

aminalaee:
Looks like I can't edit your fork,

can you try the following codecov configuration in `.codecov.yml`?

```
coverage:
  precision: 2

comment: false
```

yezz123:
I just add it, I don't know why GitHub didn't show the button when we give access to write 🤔

.codecov.yml
```diff
@@ -1 +1,4 @@
-comment: off
+coverage:
+  precision: 1
+
+comment: false
```
aminalaee, line 4:
```suggestion
coverage:
  status:
    project:
      default:
        target: auto
        threshold: 1%

comment: false

```

aminalaee (approved):
Thanks for this.

PR #29, opened 2022-01-27
add the login function

Chise1:
I tried adding the login function. Including the following:
1. A User model
2. A script to create a user
3. Reads the fields required for user-generated passwords from. Env.
4. Login html can use i18n and Chinese.

Now there has some question:
1.I don't know how to create an executable with setup.py.(I can only use poetry to create a program by `tool.poetry.scripts`)
2. Most of this update is from my old project.But it need deeply test.
3. Doc need update.

In new version,you need create a .env in project's root.And .env need this data(like this): 
SECRET_KEY=aawinjwol;egbfnjek bnl
DATABASE_URL=sqlite:///example.db

If you want to create a manager:
bash
```
sqladmin username password
```


Chise1:
May be this project need use poetry to replace requirements.txt+setup.py

sqladmin/application.py
```diff
@@ -34,13 +41,26 @@ def __init__(
...
     ) -> None:
         self.app = app
         self.engine = engine
         self.base_url = base_url
         self._model_admins: List[Type["ModelAdmin"]] = []
 
         self.templates = Jinja2Templates("templates")
+        self.templates.env.add_extension("jinja2.ext.i18n")
+        if language:
+            translation = gettext.translation(
```
aminalaee, line 54: I'd rather do this in a separate changes to keep small PR and make review easier. What do you think?

sqladmin/application.py
```diff
@@ -204,7 +258,12 @@ async def list(self, request: Request) -> Response:
 
     async def detail(self, request: Request) -> Response:
         """Detail route."""
-
+        if not check_token(request):
```
aminalaee, line 261: A `@login_required` would be easier here.

sqladmin/auth/models.py
```diff
@@ -0,0 +1,29 @@
+from sqlalchemy import Boolean, Column, Integer, String
+from sqlalchemy.ext.declarative import declarative_base
+
+from sqladmin.auth.hashers import make_password, verify_password
+
+Base = declarative_base()
+
+
+class User(Base):  # type: ignore
```
aminalaee, line 9: Actually I thought we would integrate with something like https://www.starlette.io/authentication/ . I guess it would be easier to leave these out of the admin itself. what do you think?

aminalaee (commented):
Thank you for the PR. This is a great start.

aminalaee:
> May be this project need use poetry to replace requirements.txt+setup.py

Actually I use poetry in another project but I think this seems more reasonable. Why do you think so?

sqladmin/application.py
Chise1: I'll pay attention next time.

If you think I submit too many contents this time, I can submit them separately.

sqladmin/auth/models.py
Chise1: In fact, I'm more familiar with FastAPI.

Your proposal is very good, I will look at the link you sent tomorrow and try to modify it.

aminalaee: I think they should basically be the same. Sure.

Chise1:
I think poetry is easier and cheaper.

Chise1:
I ran into a problem that the./script/build directive would run fine on my local, but not in CI/CD.

aminalaee:
> I ran into a problem that the./script/build directive would run fine on my local, but not in CI/CD.

I'll fix that separately in another PR. Thanks

sqladmin/application.py
aminalaee: that would be great. If you could focus this PR on the login functionality and do other PRs for translation and other things I'd really appreciate that.
PRs are always appreciated :)

Chise1:
I have cleared i18n code and add 'create manager' cli.

aminalaee:
This generally looks good, but to be honest I think we should probably do something like Flask-Admin does [here](https://flask-admin.readthedocs.io/en/latest/introduction/#authorization-permissions), and leave all user management, cli and things like that out of SQLAdmin. What do you think?
I don't have like a clear idea how to do that but if you feel like you want to do it, I think that approach seems reasonable.

Chise1:
The Spring Festival is coming. I want to have a rest.

You're right, the question of whether you need to introduce authentication and permissions really needs to be thought about.

I'll get back to you over the holidays.

Chise1:
The Spring Festival is coming. I want to have a rest.

You're right, the question of whether you need to introduce authentication and permissions really needs to be thought about.

I'll get back to you over the holidays.

Chise1:
The Spring Festival is coming. I want to have a rest.

You're right, the question of whether you need to introduce authentication and permissions really needs to be thought about.

I'll get back to you over the holidays.

Chise1:
Write

PR #95, opened 2022-03-18
Fix Source Code Link

baurt:
Source code link was pointing to encode/starlette. Changed to current project.


aminalaee (approved):
Thank you for the PR 👍

Issue #96, opened 2022-03-20
Add "form_columns" concept


dwreeves:
### Checklist

- [X] There are no similar issues or pull requests for this yet.

### Is your feature related to a problem? Please describe.

I am trying to emulate all the functionality of a Flask-Admin panel I made using SQLAdmin. The Flask website I'm attempting to emulate is open source and available [here](https://github.com/codeforboston/flagging). It is able to be deployed in "demo" mode to Heroku so anyone can run it. (set `FLASK_ENV=demo` as you deploy it).

So far, I am running into an issue with complicated lazy relationships. Take the following SQLModel as an example:

<img width="865" alt="image" src="https://user-images.githubusercontent.com/31971762/159168988-a68cd782-acf9-4dd2-a49a-5818b66243fe.png">

This is a SQLModel with lots of complicated relationships at the bottom, but these relationships do not need to be manually defined in the admin panel. In the case of `latest_prediction` it's resolved via `max(timestamp)`. In the case of `all_predictions`, it is a many-to-one relationship that is resolved through an associative entity (named `Reach`).

In Flask-Admin, I would work around this by explicitly defining `form_columns` and excluding these complicated fields. And for Flask-Admin, this is what I get for my `Boathouse` entity type:

<img width="1147" alt="image" src="https://user-images.githubusercontent.com/31971762/159168913-9a5ced3f-1c04-45e5-8415-de1351c71938.png">

But due to lack of a `form_columns`, I cannot make this happen in SQLAdmin. It is forcing me to include these columns, which means I effectively cannot use both SQLAdmin and these complicated fields at the same time:

![image](https://user-images.githubusercontent.com/31971762/159168961-4180cc62-6c43-4ba5-8341-ebade63fabc5.png)

### Describe the solution you would like.

Flask-Admin has a concept called `form_columns` for its `ModelView` class. The `form_columns` attribute controls which columns that WTForms creates fields for.

This is a good catchall solution for many problems because it lets users avoid any issues relating to fields that do not need to be assigned as part of the initialization of an object (e.g. optional fields, fields with defaults, complicated relationships, and more).

Flask-Admin implements `form_columns` here, primarily: https://github.com/flask-admin/flask-admin/blob/7cff9c742d44d42a8d3495c73a6d71381c796396/flask_admin/contrib/sqla/form.py#L94

---

[...]

PR #97, opened 2022-03-20
Add form-specific functionality to ModelAdmin

dwreeves:
I added the following attributes to `ModelAdmin`:

* `form`
* `form_base_class`
* `form_args`
* `form_columns`
* `form_excluded_columns`
* `form_overrides`

All of these attributes emulate their respective functionality in the Flask-Admin API: https://flask-admin.readthedocs.io/en/latest/api/mod_model/

This PR addresses issue #96.

--

Note that I abstracted out the `get_{x}_columns()` functionality in `ModelAdmin`. Let me know if you want me to revert this back to being repetitive.

---

~~As of writing, there is one thing currently missing from my PR: the scaffolding does _not_ make use of `column_labels`. This is something I will hope to finish up soon!~~

Let me know what other changes you want me to make.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -1,9 +1,11 @@
 from enum import Enum
 from typing import (
     Any,
+    Callable,
     ClassVar,
     Dict,
     List,
+    Optional,
     Sequence,
     Tuple,
     Type,
@@ -71,6 +73,9 @@ class ModelAdminMeta(type):
         cls.icon = attrs.get("icon")
 
         mcls._check_conflicting_options(["column_list", "column_exclude_list"], attrs)
+        mcls._check_conflicting_options(
+            ["form_columns", "form_excluded_columns"], attrs
+        )
         mcls._check_conflicting_options(
             ["column_details_list", "column_details_exclude_list"], attrs
         )
@@ -283,6 +288,63 @@ class ModelAdmin(BaseModelAdmin, metaclass=ModelAdminMeta):
     edit_template: ClassVar[str] = "edit.html"
     """Edit view template. Default is `edit.html`."""
 
+    # Form
+    form: ClassVar[Optional[Type[Form]]] = None
+    """Form class.
+    Override if you want to use custom form for your model.
+    Will completely disable form scaffolding functionality.
+
+    ???+ example
+        ```python
+        class MyForm(Form):
+            name = StringField('Name')
+
+        class MyModelAdmin(ModelAdmin, model=User):
+            form = MyForm
+        ```
+    """
+
+    form_base_class: ClassVar[Optional[Type[Form]]] = Form
+    """Base form class.
+    Will be used by form scaffolding function when creating model form.
+    Useful if you want to have custom constructor or override some fields.
+
+    ???+ example
+        ```python
+        class MyBaseForm(Form):
+            def do_something(self):
+                pass
+
+        class MyModelAdmin(ModelAdmin, model=User):
+            form_base_class = MyBaseForm
+        ```
+    """
+
+    form_columns: ClassVar[Sequence[Union[str, InstrumentedAttribute]]] = []
+    """List of columns to include in the form.
+    Columns can either be string names or SQLAlchemy columns.
+
+    ???+ note
+        By default all columns of Model are included in the form.
+
+    ???+ example
+        ```python
+        class UserAdmin(ModelAdmin, model=User):
+            form_columns = [User.name, User.mail]
+        ```
+    """
+
+    form_excluded_columns: ClassVar[Sequence[Union[str, InstrumentedAttribute]]] = []
+    """List of columns to exclude from the form.
+    Columns can either be string names or SQLAlchemy columns.
+
+    ???+ example
+        ```python
+        class UserAdmin(ModelAdmin, model=User):
+            form_excluded_columns = [User.id]
+        ```
+    """
+
     def __init__(self) -> None:
         self._column_labels = self.get_column_labels()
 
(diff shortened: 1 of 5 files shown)


dwreeves:
I added support for `form_args`, and I also pass column labels into the form.

The implementation is... not the best. I still largely retain the behavior you have implemented, which is to loop over the `str` identifiers for the columns rather than the actual Columns. I think the "Form" part of the code is starting to outgrow mere `str`s. 😬 There should be some refactoring to make this work more with Column objects rather than strings, and a few other things. Hard to explain but I think you'll see what I mean when you peek at my code.

dwreeves:
Just added `form_overrides` as well.

Again, this will definitely need to be refactored in the future (I already implemented a little bit of refactoring of something that was there, i.e. the `get_converter()` method).

Regardless, it should be emulating the behavior of the Flask-Admin API.

aminalaee:
Hey, thanks a lot for this. It generally looks very good to me.
I wish you didn't add all of them in one PR as it makes the review a bit harder. But I'll try my best to review it today 👍

sqladmin/models.py
```diff
@@ -445,43 +532,62 @@ def get_model_attr(
...
 
+    def _build_column_list(
+        self,
+        include: Optional[Sequence[Union[str, InstrumentedAttribute]]] = None,
+        exclude: Optional[Sequence[Union[str, InstrumentedAttribute]]] = None,
+        default: Callable[[], List[Column]] = None,
+    ) -> List[Tuple[str, Column]]:
+        """Many situations in the code call for constructing a list of columns.
+        This function generalizes this pattern for any sequence of inclusions
+        or exclusions.
```
aminalaee, line 543:
```suggestion
        """This function generalizes constructing a list of columns
        for any sequence of inclusions or exclusions.
```

sqladmin/models.py
```diff
@@ -283,6 +288,88 @@ class UserAdmin(ModelAdmin, model=User):
...
+        ```python
+        class MyForm(Form):
+            name = StringField('Name')
+
+        class MyModelAdmin(ModelAdmin, model=User):
+            form = MyForm
+        ```
+    """
+
+    form_base_class: ClassVar[Optional[Type[Form]]] = Form
```
aminalaee, line 307: Do we need the Optional here?

sqladmin/models.py
```diff
@@ -283,6 +288,88 @@ class UserAdmin(ModelAdmin, model=User):
     edit_template: ClassVar[str] = "edit.html"
     """Edit view template. Default is `edit.html`."""
 
+    # Form
+    form: ClassVar[Optional[Type[Form]]] = None
```
aminalaee, line 292: I think technically you're right to have Optional here, but since we're not using it as arguments it won't be checked? What do you think?

sqladmin/models.py
```diff
@@ -283,6 +288,88 @@ class UserAdmin(ModelAdmin, model=User):
...
+        class MyBaseForm(Form):
+            def do_something(self):
+                pass
+
+        class MyModelAdmin(ModelAdmin, model=User):
+            form_base_class = MyBaseForm
+        ```
+    """
+
+    form_args: ClassVar[Optional[Dict[str, Dict[str, Any]]]] = None
```
aminalaee, line 323: And is Optional required here? Can we default to `{}` as this is not a function argument?

sqladmin/models.py
```diff
@@ -283,6 +288,88 @@ class UserAdmin(ModelAdmin, model=User):
...
+    Columns can either be string names or SQLAlchemy columns.
+
+    ???+ example
+        ```python
+        class UserAdmin(ModelAdmin, model=User):
+            form_excluded_columns = [User.id]
+        ```
+    """
+
+    form_overrides: ClassVar[Optional[Dict[str, Type[Field]]]] = None
```
aminalaee, line 363: And here?

sqladmin/models.py
```diff
@@ -525,7 +631,17 @@ async def update_model(self, pk: Any, data: Dict[str, Any]) -> None:
             await anyio.to_thread.run_sync(self._update_modeL_sync, pk, data)
 
     async def scaffold_form(self) -> Type[Form]:
-        return await get_model_form(model=self.model, engine=self.engine)
+        if self.form is not None:
+            return self.form
```
aminalaee, line 635: Coverage missing here.

aminalaee: I don't have the right permissions, I tried to add the test:

sqladmin/models.py
dwreeves: We can avoid the `Optional` here, yes.

sqladmin/models.py
dwreeves: So basically, all the other Optionals I think are fine to remove. But I don't think this one should be removed.

sqladmin/models.py
dwreeves: I don't see what test you were trying to add, but I added one called `test_form_override_scaffold` inside `test_forms.py`. Good catch on lack of coverage.

aminalaee: Yes, that was exactly it.

dwreeves:
✅ I removed most of the `Optional` class vars. I do think that `form` should remain optional (based on the specification outlined in the Flask-Admin API), but if you have an idea of an alternate implementation, let me know.

✅ I added coverage for the override of the scaffolding.

sqladmin/models.py
aminalaee: Yes, you're right.

aminalaee (approved):
Thanks a lot for the great work @dwreeves 🎉

aminalaee:
> Woo, all checks passed! 🎉
> 
> Thanks for all the hard work you've put into this library so far. It was lots of fun to help contribute. This is a really good implementation of an admin panel framework.
> 
> I plan on making tons of use of this library for my own projects. You might see some more PRs from me in the near future, especially as it pertains to porting more things over from Flask-Admin.

I really appreciate the effort and all sorts of contributions are welcome :)

I'm glad you'll use it and contribute to make it better.

PR #101, opened 2022-03-23
[Feature] Add csv export support

dwreeves:
I added csv export support.

For multiple export types, the GUI is ready, but the backend implementation of other types is not.

Also, I made it so `can_create = False` alters the UI and hides the create button.

Visual demo here: https://www.youtube.com/watch?v=Vk4luHsFgrQ

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -334,10 +334,9 @@ class ModelAdmin(BaseModelAdmin, metaclass=ModelAdminMeta):
     Currently only `csv` is supported.
     """
 
-    export_max_rows: ClassVar[Optional[int]] = 0
+    export_max_rows: ClassVar[int] = 0
     """Maximum number of rows allowed for export.
-
-    Unlimited by default. Uses `page_size` if set to `None`.
+    Unlimited by default.
     """
 
     # Form
@@ -660,7 +659,7 @@ class ModelAdmin(BaseModelAdmin, metaclass=ModelAdminMeta):
         return self._build_column_list(
             include=columns,
             exclude=excluded_columns,
-            default=lambda: [getattr(self.model, self.pk_column.name).prop],
+            default=lambda: self._list_columns,
         )
 
     def get_column_labels(self) -> Dict[Column, str]:
@@ -757,7 +756,6 @@ class ModelAdmin(BaseModelAdmin, metaclass=ModelAdminMeta):
         self,
         data: List[Any],
     ) -> StreamingResponse:
-
         def generate(writer: Writer) -> Generator[List[str], None, None]:
             # Append the column titles at the beginning
             titles = [c[0] for c in self._export_attrs]
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -1,4 +1,3 @@
-import csv
 import time
 from enum import Enum
 from typing import (
@@ -6,6 +5,7 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
+    Generator,
     List,
     Optional,
     Sequence,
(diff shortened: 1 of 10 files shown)


dwreeves:
Looks like I need a little more coverage... I'll get back to this tomorrow!

sqladmin/models.py
```diff
@@ -288,6 +304,40 @@ class UserAdmin(ModelAdmin, model=User):
...
+    export_types: ClassVar[List[str]] = ["csv"]
+    """A list of available export filetypes.
+    Currently only `csv` is supported.
+    """
+
+    export_max_rows: ClassVar[Optional[int]] = 0
+    """Maximum number of rows allowed for export.
+
+    Unlimited by default. Uses `page_size` if set to `None`.
+    """
```
aminalaee, line 339: What do you think if we don't follow the Flask-Admin behaviour here? I don't think it would be a problem here.

```suggestion
    export_max_rows: ClassVar[int] = 0
    """Maximum number of rows allowed for export.
    Unlimited by default.
    """
```

sqladmin/models.py
```diff
@@ -588,6 +647,20 @@ def get_form_columns(self) -> List[Tuple[str, Column]]:
             default=self.get_model_attributes,
         )
 
+    def get_export_columns(self) -> List[Tuple[str, Column]]:
+        """Get list of columns to export."""
+
+        columns = getattr(self, "column_export_list", None)
+        excluded_columns = getattr(self, "column_export_exclude_list", None)
+        if not columns and not excluded_columns:
+            return self.get_list_columns()
```
aminalaee, line 656:
```suggestion
```

sqladmin/models.py
```diff
@@ -588,6 +647,20 @@ def get_form_columns(self) -> List[Tuple[str, Column]]:
...
+
+        return self._build_column_list(
+            include=columns,
+            exclude=excluded_columns,
+            default=lambda: [getattr(self.model, self.pk_column.name).prop],
```
aminalaee, line 661:
```suggestion
            default=lambda: self._list_columns,
```

sqladmin/models.py
```diff
@@ -662,3 +735,63 @@ class UserAdmin(ModelAdmin, model=User):
...
         ]
         return ", ".join(field_names)
+
+    def get_export_name(self, export_type: str) -> str:
+        """The file name when exporting."""
+        filename = "%s_%s.%s" % (
+            self.name,
+            time.strftime("%Y-%m-%d_%H-%M-%S"),
+            export_type,
+        )
```
aminalaee, line 745:
```suggestion
        filename = f"{self.name}_{time.strftime('%Y-%m-%d_%H-%M-%S')}.{export_type}"
```

sqladmin/models.py
```diff
@@ -662,3 +735,63 @@ class UserAdmin(ModelAdmin, model=User):
...
+            return self._export_csv(data)
+        else:
+            raise NotImplementedError("Only export_type='csv' is implemented.")
+
+    def _export_csv(
+        self,
+        data: List[Any],
+    ) -> StreamingResponse:
+        # https://docs.djangoproject.com/en/1.8/howto/outputting-csv/
+        class Echo(object):
```
aminalaee, line 763: Maybe this can go into helpers.py?

aminalaee (commented):
Thanks a lot for the PR, great work 👍

sqladmin/models.py
dwreeves: (I opted to not do it because it relies on execution order. But beneficially, it also more strongly couples the default behavior to copying list columns. Bit of a trade-off...)

sqladmin/models.py
dwreeves: I think Flask-Admin's default behavior is very weird-- It's odd that `0` is default and means infinite, and `None` is a choice and has a completely different behavior. I think deviating is justifiable.

sqladmin/models.py
dwreeves: I tried to separate out as much of the API logic from the parsing logic with a function called `stream_to_csv`.

Issue #104, opened 2022-03-23
[Bug] Docs are rendering docstrings from `wtforms.Form` and not `form_base_class`.


dwreeves:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

This is not intended:

<img width="1420" alt="image" src="https://user-images.githubusercontent.com/31971762/159769748-c540616a-846e-45d9-85cc-4265b01e0cd8.png">

### Steps to reproduce the bug

Deploy the docs.

### Expected behavior

The docstring should be from the file `sqladmin/models.py`:

<img width="647" alt="image" src="https://user-images.githubusercontent.com/31971762/159769841-22a6a9d7-ecad-410f-9328-eb894f284b36.png">

### Environment

Github Actions

PR #105, opened 2022-03-23
Update models.py

aminalaee:


dwreeves:
In reference to #104-- I think the problem with the form base class docs is a problem with `mkdocstrings`. The typing I did was fine.

I fixed #104 in my local environment by adding `mkdocstrings-python==0.6.6` to `requirements.txt`. Long story short, you weren't specifying a handler for mkdocstrings so it was defaulting to the "legacy" handler. However, the package they developed called [griffe](https://github.com/mkdocstrings/griffe) properly identifies the docstring. (You can see this by `pip install griffe` then `griffe sqladmin` and you'll see that `form_base_class`'s docstring is properly evaluated with griffe. This is not the case for the legacy parsing that mkdocstrings used.)

aminalaee:
Thank you, yes I was just experimenting with the docstrings, I didn't really expect any result here.

PR #107, opened 2022-03-24
Fix Export unlimited rows

aminalaee:
Right now if you don't specify `export_max_rows` then only 10 rows are exported. An example here: https://python-sqladmin.herokuapp.com/admin/user/list

The reason this happened is that we rely on the pagination and pagination will do a few checks before setting page and page_size options and won't allow unlimited rows.

I added a different query for this behaviour.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -375,5 +375,5 @@ class Admin(BaseAdminView):
         export_type = request.path_params["export_type"]
 
         model_admin = self._find_model_admin(identity)
-        pagination = await model_admin.list(1, model_admin.export_max_rows)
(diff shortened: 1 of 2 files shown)


aminalaee:
cc @dwreeves

dwreeves:
Oh woops! Good catch. I'll test this locally soon, and also we should add a test case for N>10 rows.

dwreeves:
LGTM. Your fix works when I run it. 👍

I added test coverage for `export_max_rows` here: #108

dwreeves (approved):

aminalaee:
Docs are failing, nothing related here.

Issue #113, opened 2022-03-29
Expose Starlette options


tr11:
### Checklist

- [X] There are no similar issues or pull requests for this yet.

### Is your feature related to a problem? Please describe.

It'd be great to expose the `middleware` and `debug` parameters to pass to the Starlette admin app. This would allow, for example, to create a middleware that checks for an "admin" session cookie and redirect the user in case the cookie doesn't exist.  The debug option is useful when developing.

aminalaee:
Thank you for the suggestion and PR, this is already merged.

PR #114, opened 2022-03-29
Expose Starlette middlewares and debug to the Admin

tr11:
Adds the `middleware` and `debug` parameters as mentioned in #113.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List, Type, Union
+from typing import TYPE_CHECKING, List, Sequence, Type, Union, Optional, Dict, Any
 
 from jinja2 import ChoiceLoader, FileSystemLoader, PackageLoader
 from sqlalchemy.engine import Engine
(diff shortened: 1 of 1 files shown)


sqladmin/application.py
```diff
@@ -156,6 +157,8 @@ def __init__(
         base_url: str = "/admin",
         title: str = "Admin",
         logo_url: str = None,
+        middleware: Sequence[Middleware] = None,
```
aminalaee, line 160: I know that Starlette expects `middleware` keyword, but since this is a sequence, I think it's better to switch this to `middlewares` on the SQLAdmin side.
```suggestion
        middlewares: Sequence[Middleware] = None,
```

sqladmin/application.py
```diff
@@ -212,6 +215,8 @@ def http_exception(request: Request, exc: Exception) -> Response:
                 ),
             ],
             exception_handlers={HTTPException: http_exception},
+            middleware=middleware,
```
aminalaee, line 218:
```suggestion
            middleware=middlewares,
```

aminalaee (approved):
Thank you for the PR 🎉

PR #115, opened 2022-03-29
Refactor `forms.py` (Work in progress)

dwreeves:
## Objectives

This PR is a refactor of `forms.py`.

The immediate objectives of this refactor are listed in the section titled "Code Changes".

The longer-term objective of this refactor is to move toward using `AdminAttribute`s internally. A lot of the code right now is hard to manage because there is no concept that relates a single SQLA/wtforms field/column/property. Ultimately, a `List[AdminAttribute]` should be generated by the `ModelAdmin()` instance, instead of inside `get_model_form()`.

In my opinion, more and more parsing/app logic should move toward `AdminAttribute` internals, and then the rest of the code can focus more on the core framework logic. Right now, a lot of parsing, and generation of defaults, is being handled at multiple parts of the code. `AdminAttribute` can be a single spot for all the "ugly but necessary" stuff, and the rest of the code (the parts we want to actually add features) will be cleaner as a result.

### Consistent typing:

- Create a `ConverterCallback` type with a sensible signature. (In this case, pulled from Flask-Admin's pattern.)
- Apply the signature to decorated `@converts(...)` methods.
- wtforms has very weird typing (or rather, a lack of typing). I added a type for the wtforms validators. I got the type right (check `wtforms/validators.py`), but it's unclear whether this is appropriate.

#### Converter Callbacks

The main purpose of the renaming in this refactor is to enforce the distinction between a **converter** and a **converter callback**. The class is called `ModelConverter`, which implies that a `ModelConverter()` object is the _converter_. But, in the code, `get_converter()` does not get a converter instance, it gets a callable that is not a converter _per se_ (based on the convention that a `ModelConverter` is a converter.

To have consistent naming, the function should be called a "converter callback," to distinguish it from the converter class. All references to the decorated functions should then make clear that these are "converter callbacks" and not mere "converters."

- `ModelConverterBase.get_converter()` --> `ModelConverterBase.get_callback()`
- `ModelConverterBase.converters` --> `ModelConverterBase._callbacks`
- Create a type called `ConverterCallback` to reinforce both the concept and the signature.

#### Property vs attribute

`mapper.attrs` misleadingly returns objects of type `Union[ColumnProperty, RelationshipProperty]`. I think we should move toward referring to objects with this typing as `prop`s. In the function `get_model_form`, I adhere to this behavior. (Although I don't make any additional changes in the rest of the code.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ sqlmodel==0.0.6
 
 # Tests
 autoflake==1.4
-black==22.1.0
+black==22.3.0
 coverage==6.3.2
 flake8==4.0.1
 httpx==0.22.0
--- a/sqladmin/exceptions.py
+++ b/sqladmin/exceptions.py
@@ -1,6 +1,14 @@
-class InvalidModelError(Exception):
+class SQLAdminException(Exception):
     pass
 
 
-class InvalidColumnError(Exception):
+class InvalidModelError(SQLAdminException):
+    pass
+
+
+class InvalidColumnError(SQLAdminException):
+    pass
+
+
+class NoConverterFound(SQLAdminException):
     pass
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -1,13 +1,20 @@
 import inspect
 from enum import Enum
-from typing import Any, Callable, Dict, Optional, Sequence, Type, Union, no_type_check
+from typing import Any, Callable, Dict, List, Optional, Sequence, Type, TypeVar, Union
 
 import anyio
 from sqlalchemy import inspect as sqlalchemy_inspect, select
 from sqlalchemy.engine import Engine
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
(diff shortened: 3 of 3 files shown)


dwreeves:
It's possible even if this is good enough that it should be merged into a `dev` branch.

I don't think anyone is going that deep into this project's internals yet, but it is possibly breaking of existing workflows.

aminalaee:
Thanks for the effort @dwreeves . I think the refactoring is a good idea, but I suggest to split it into multiple PRs so that we can review this more quickly. Do you think that's possible?

dwreeves:
I'll think about it, but I'm having a hard time figuring out what I should split out of this PR. This was as atomic as I could get it; I only touched `forms.py` and deliberately did not touch `models.py`.

aminalaee:
I think the titles you mentioned make sense, so maybe start with the renaming callbscks, then typing improvements and then property and attrs. Does that make sense?

dwreeves:
Sounds good!

PR #116, opened 2022-03-30
Function signature typing, and renames

dwreeves:
I broke out the typing/renames from #115 into a separate PR.

### DIFF
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -1,6 +1,17 @@
 import inspect
 from enum import Enum
-from typing import Any, Callable, Dict, Optional, Sequence, Type, Union, no_type_check
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Type,
+    TypeVar,
+    Union,
+    no_type_check,
+)
 
 import anyio
 from sqlalchemy import inspect as sqlalchemy_inspect, select
@@ -8,6 +19,7 @@ from sqlalchemy.engine import Engine
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 from sqlalchemy.orm import ColumnProperty, Mapper, RelationshipProperty, Session
 from sqlalchemy.sql.schema import Column
+from typing_extensions import Protocol
 from wtforms import (
     BooleanField,
     DateField,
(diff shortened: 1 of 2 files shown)


sqladmin/forms.py
```diff
@@ -161,107 +199,169 @@ async def convert(
...
-    @classmethod
-    def _string_common(cls, column: Column, field_args: Dict, **kwargs: Any) -> None:
+    @staticmethod
+    def _string_common(prop: ColumnProperty) -> List[Validator]:
+        li = []
+        column: Column = prop.columns[0]
         if isinstance(column.type.length, int) and column.type.length:
-            field_args["validators"].append(validators.Length(max=column.type.length))
+            li.append(validators.Length(max=column.type.length))
+        return li
```
aminalaee, line 246: Isn't this always a list with one element?

sqladmin/forms.py
```diff
@@ -161,107 +199,169 @@ async def convert(
...
-        field_args.setdefault("places", None)
-        return DecimalField(**field_args)
+        kwargs.setdefault("places", None)
+        return DecimalField(**kwargs)
 
     # @converts("dialects.mysql.types.YEAR", "dialects.mysql.base.YEAR")
-    # def conv_MSYear(self, field_args: Dict, **kwargs: Any) -> Field:
-    #     field_args["validators"].append(validators.NumberRange(min=1901, max=2155))
-    #     return StringField(**field_args)
+    # def conv_MSYear(
```
aminalaee, line 314: I'd rather leave this out for now.

sqladmin/forms.py
```diff
@@ -23,37 +35,63 @@
...
-                    converters[classname] = obj
+                    callbacks[classname] = obj
 
-        self.converters = converters
+        self._callbacks = callbacks
 
-    def get_converter(
-        self, prop: Union[ColumnProperty, RelationshipProperty]
-    ) -> Callable:
+    def get_callback(
```
aminalaee, line 90: What's the reason for renaming converter to callback? Is it unclear?

sqladmin/forms.py
```diff
@@ -23,37 +35,63 @@
...
+        ...  # pragma: no cover
+
+
+class ConverterCallback(Protocol):
+    def __call__(
+        self,
+        model: type,
+        prop: Union[ColumnProperty, RelationshipProperty],
+        kwargs: Dict[str, Any],
+    ) -> UnboundField:
```
aminalaee, line 56: I will need to read the mypy docs again about this. I think this PR needs to be rebased.

sqladmin/forms.py
dwreeves: The list can also be empty.

There are a few ways to write this snippet; I refactored because I thought it'd be simpler if the function did not rely on side effects.

sqladmin/forms.py
dwreeves: Right now "converter" refers to both the name of the class and the name of the methods within the class. The name "converter" currently does two things. But by distinguishing between converters (the class) and the _callbacks_ (the methods), we have 1 name for 1 concept each, rather than 1 name referring to 2 concepts.

We can potentially get the code to the point where we can expose this part of the API, and I think it is appropriate, if you want to do that, to use good and precise names.

dwreeves: One thing I should have considered-- is "callback" even the right terminology? I think a callback is more like, a function passed in an argument. It's tenuously a callback because it's decorated. So, I'll revert it back to `converters` and `get_converter`. And I will rename `ConverterCallback` to `ConverterCallable` for the protoocol.

sqladmin/forms.py
aminalaee: I think we can avoid changing this. we can update it when adding it. Does that make sense?

sqladmin/forms.py
aminalaee: Yes I agree, this is a pretty old approach, I honestly took it from WTForms and didn't go back to re-factoring it.

sqladmin/forms.py
dwreeves: Sure

sqladmin/forms.py
dwreeves: I rebased the PR.

aminalaee (approved):
Thanks a lot @dwreeves this looks good 👍

PR #129, opened 2022-04-05
Delay call to `self.get_converter`

lovetoburnswhen:
- This enables users who pass in `form_overrides` to bypass `self.get_converter()`, which can still break on custom/unsupported types (eg `sqlalchemy_utils.types.ChoiceType`)


dwreeves (approved):

aminalaee (approved):
Thank you for the PR 👍

PR #133, opened 2022-04-06
Improve typing of models.py attributes

aminalaee:


sqladmin/models.py
```diff
@@ -185,7 +185,7 @@ class UserAdmin(ModelAdmin, model=User):
     """
 
     # List page
-    column_list: ClassVar[Sequence[Union[str, InstrumentedAttribute]]] = []
+    column_list: ClassVar[Sequence[Union[str, Column, RelationshipProperty]]] = []
```
dwreeves, line 188: I don't think this is the correct typing. The original typing worked fine.

aminalaee: Hmmm..you're right.
It's strange though I fixed this because of some mypy issues complaining it's not InstrumentedAttribute.
Thanks for the review.

PR #140, opened 2022-04-24
Add column_formatters

### LINKED ISSUE #131: Ability to choose format of datetimes in list/table view
### Checklist

Currently the values for datetime columns seem to be rendered as ISO-8601 with millisecond precision which is unnecessary (eg `2022-04-04 22:43:30.027151`)

It'd be great if we had a way to pass some sort of custom formatter, ideally a callable that takes the DB value and returns the value that should be passed to the renderer.

_No response_

### Additional context

_No response_

skarrok:
Closes #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unrealeased
+
+### Added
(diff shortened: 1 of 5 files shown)


skarrok:
Hey, thanks for the review!
I wanted to add proper description to PR and maybe add couple more examples in docs, like with datetime in linked issue.
But got caught up in work and stuff.

Also was thinking maybe we should pass more information/context to formatter function, but I guess it can be added later if needed without breaking compatibility.

So if this already looks good to you I'm all in for merging as is :)

aminalaee (approved):
Thank you for the PR @skarrok 🎉

PR #149, opened 2022-05-10
Add form_ajax_refs

aminalaee:


sqladmin/fields.py
```diff
@@ -425,3 +427,92 @@ def pre_validate(self, form: Form) -> None:
...
+        super().__init__(label, validators, **kwargs)
+        self.loader = loader
+
+        self.allow_blank = allow_blank
+        self.blank_text = blank_text
+
+    @property
+    def data(self) -> Any:
+        if self._formdata:
+            self.data = self._formdata
```
aminalaee, line 458: The most important change needed is that here we're setting the PK of that object inside in the form, which as part of the design we expect to set the SQLAlchemy object, not the PK only.

the challenge is that it's not possible to query the async db in the sync WTForm method. 
The only way I could think of is to handle this when inserting the object.

inside the insert we can check if this attribute name is a relationship and if it is, query based on that PK to get the SQLAlchemy object with it. It's a bit hackish but I'll look for other approaches.

aminalaee: It would be great if @dwreeves could take a look.

dwreeves: Sorry, started a new job this week. I will take a look this weekend!

dwreeves (commented):
Wow, lots of big changes to the code!

I loaded up the `flask-admin/examples/sqla` demo app to see how `form_ajax_refs` is supposed to work in Flask-Admin.

I'm still trying to get a demo of this working properly. I have my test FastAPI+Sqladmin app running with ajax refs:

image

Right now the WTForms part is working, but the actual query isn't filtering. I am using the synchronous engine. I'll play around with this a bit more throughout the week.

Hopefully this is a good example of the problem I am running into. I created 26 entries with the following:

```python
    reaches = [
        Reach(id=i+1, name=letter)
        for i, letter
        in enumerate("abcdefghijklmnopqrstuvwxyz")
    ]
```

And this is what I see when I try to search for `z`:

image

I spent a couple hours on this today (looking through this code + Flask-Admin's). I'm going to take a break and come back to this throughout the week as I have time.

Looks promising, though, and can't wait for this to be in!

sqladmin/statics/js/main.js
```diff
@@ -35,15 +35,33 @@ $(document).on('click','#search-button',function() {
...
+  $(obj).select2({
+    placeholder: 'Search for a repository',
+    minimumInputLength: 1,
+    ajax: {
+      url: $(this).data("url"),
+      dataType: 'json',
+      data: function (params) {
+        var query = {
+          name: $(this).attr("name"),
+          limit: 20,
```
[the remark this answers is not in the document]
aminalaee: @dwreeves Thanks, I'll need to complete the actual query, I still haven't figured out how to do that.

PR #150, opened 2022-05-10
Handling for sqlalchemy_utils EmailType and IPAddressType

### LINKED ISSUE #148: Used Sqlalchemy_utils types
### Checklist

I want to be able to use model types provided by sqlalchemy_utils such as EmailType, IPAddressType, and UUIDType.

I would like the admin page to work without crashing when they are used (just treating them all as strings would work fine for me).

I could convert them all to strings in my model, but that would take away a lot of functionality.

### Additional context

_No response_

colin99d:
Currently if an EmailType or IPAddressType from sqlalchemy_utils is used it causes the entire admin to crash. Improved this feature by adding integrating validators from wtforms.  Fixes #148 

Note: I did need to add a dependency, email-validators. See the following stack overflow post for why this was necessary: https://stackoverflow.com/questions/61356834/wtforms-install-email-validator-for-email-validation-support


aminalaee:
hey, thanks a lot for the PR.

Just a thought, why should the `email-validators` be a requirement in SQLAdmin? It is imported at runtime by WTForm so it can be installed independently.

colin99d:
> hey, thanks a lot for the PR.
> 
> Just a thought, why should the `email-validators` be a requirement in SQLAdmin? It is imported at runtime by WTForm so it can be installed independently.

Great point! Changes made.

aminalaee:
It's only missing test coverage, you can add one here, specifically for sqlalchemy_utils.
I think you need to add sqlalchemy-utils to the requirements.txt (dev dependencies) to test it.

https://github.com/aminalaee/sqladmin/blob/f5cd0f34130ab769cd3e49a3d8c69983eb304dd8/tests/test_forms.py#L149-L159

colin99d:
Got the tests added in. Also had to add email-validators as a dependency for tests to run.

aminalaee (approved):
Thanks a lot for the PR 🎉

PR #177, opened 2022-06-15
Use __class__ when impl is not callable

okapies:
Fixes aminalaee/sqladmin#176.


aminalaee:
@okapies it's ok. I think it might be missing the testsl coverage now.

okapies:
We cannot test it at the moment because there is no way to configure a custom converter directly (https://github.com/aminalaee/sqladmin/issues/93#issuecomment-1156103788) from outside of `ModelConverterBase`.

aminalaee:
@okapies I have to close this PR and use the other one which also includes the test, I hope you understand this and thank you for the PR.

okapies:
Thanks for the quick fix!

Issue #167, opened 2022-06-03
View Template: AttributeError: 'NoneType' object has no attribute 'id'


TomerHeber:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

I'm getting a 500 error when accessing details.

### Steps to reproduce the bug

1. Select the model in /admin/.
2. Click view on any item. 

### Expected behavior

Load the view template.

### Actual behavior

500 error.
Exception.

### Debugging material

INFO:     127.0.0.1:63769 - "GET /file/details/c09f63a4-7093-45f6-8498-17bbcd7e40e6 HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 372, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/fastapi/applications.py", line 269, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 93, in __call__
    raise exc
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    await self.app(scope, receive, sender)
  File "/Users/theber/code/college-toolkit/.venv/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
[...]

PR #182, opened 2022-06-17
Fix to handle one-to-one relationship between models

### LINKED ISSUE #180: `ModelConverter` fails to handle one-to-one relationship
### Checklist

### Describe the bug

`get_converter()` in `ModelConverterBase` selects `QuerySelectMultipleField` instead of `QuerySelectField` for a column with `ONETOMANY` direction even if its `uselist` is `False` (i.e. the column is defined as [*one-to-one* relationship](https://docs.sqlalchemy.org/en/14/orm/relationship_api.html#sqlalchemy.orm.relationship.params.uselist)).

> If a scalar is desired where normally a list would be present, such as a bi-directional one-to-one relationship, set [`relationship.uselist`](https://docs.sqlalchemy.org/en/14/orm/relationship_api.html#sqlalchemy.orm.relationship.params.uselist) to False.

Because of that, the following code tries to iterate over the non-iterable column data and fails with Internal Server Error in edit pages:

```python
from sqlalchemy import Column, ForeignKey, Integer, create_engine
from sqlalchemy.ext.declarative import declarative_base

Base = declarative_base()
engine = create_engine(...)


class User(Base):
    __tablename__ = "user"

    address: "Address" = relationship(
        "Address",
        back_populates="user",
        uselist=False,  # indicates it is one-to-one relationship
    )


class Address(Base):
    __tablename__ = "address"

    user_id = Column(Integer, ForeignKey("user.id"), unique=True, nullable=False)
    user: User = relationship("User", back_populates="address")


```

```python
from fastapi import FastAPI
from sqladmin import Admin, ModelAdmin

app = FastAPI()
admin = Admin(app, engine)


class UserAdmin(ModelAdmin, model=User):
    column_list = [User.id, User.address]


admin.register_model(UserAdmin)
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

```python
Traceback (most recent call last):
  ...
  File "/home/myproject/.venv/lib/python3.9/site-packages/sqladmin/templates/edit.html", line 18, in block 'content'
    {{ field() }}
  File "/home/myproject/.venv/lib/python3.9/site-packages/wtforms/fields/core.py", line 172, in __call__
    return self.meta.render_field(self, kwargs)
  File "/home/myproject/.venv/lib/python3.9/site-packages/wtforms/meta.py", line 64, in render_field
    return field.widget(field, **render_kw)
  File "/home/myproject/.venv/lib/python3.9/site-packages/wtforms/widgets/core.py", line 351, in __call__
    for val, label, selected in field.iter_choices():
  File "/home/myproject/.venv/lib/python3.9/site-packages/sqladmin/fields.py", line 411, in iter_choices
    primary_keys = [str(inspect(m).identity[0]) for m in self.data]
TypeError: 'Address' object is not iterable
```

### Environment

- Ubuntu 20.04.4 LTS
- Python 3.9.12
- SQLAdmin 0.1.9

### Additional context
Here is a workaround for this issue:

```python
class UserAdmin(ModelAdmin, model=User):
    ...
    form_overrides = {
        "address": QuerySelectField,  # workaround for aminalaee/sqladmin#180
    }
```

okapies:
Fixes #180.

## Note
- It seems `lazy="joined"` mess things up in async mode

```python
    addresses = relationship("Address", back_populates="user", lazy="joined")
```

```python
.venv/lib/python3.8/site-packages/starlette/routing.py:65: in app
    response = await func(request)
sqladmin/application.py:313: in create
    Form = await model_admin.scaffold_form()
sqladmin/models.py:842: in scaffold_form
    return await get_model_form(
sqladmin/forms.py:426: in get_model_form
    field = await converter.convert(
sqladmin/forms.py:222: in convert
    kwargs = await self._prepare_kwargs(
sqladmin/forms.py:149: in _prepare_kwargs
    "object_list", await self._prepare_object_list(prop, engine)
sqladmin/forms.py:173: in _prepare_object_list
    (str(self.get_pk(obj, pk)), obj) for obj in objects.scalars().all()
.venv/lib/python3.8/site-packages/sqlalchemy/engine/result.py:1371: in all
    return self._allrows()
.venv/lib/python3.8/site-packages/sqlalchemy/engine/result.py:412: in _allrows
    for made_row, sig_row in [
.venv/lib/python3.8/site-packages/sqlalchemy/engine/result.py:415: in <listcomp>
    strategy(made_row) if strategy else made_row,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = <tests.test_admin_async.User object at 0x7fd83bfe7b20>

    def require_unique(obj):
>       raise sa_exc.InvalidRequestError(
            "The unique() method must be invoked on this Result, "
            "as it contains results that include joined eager loads "
            "against collections"
        )
E       sqlalchemy.exc.InvalidRequestError: The unique() method must be invoked on this Result, as it contains results that include joined eager loads against collections
```

- I got the following warnings when running tests. I think sharing `session` between tests should be avoided especially in async mode

```python
tests/test_admin_async.py::test_update_submit_form
tests/test_admin_async.py::test_searchable_list
  sys:1: SAWarning: Identity map already had an identity for (<class 'tests.test_admin_async.User'>, (1,), None), replacing it with newly flushed object.   Are there load operations occurring inside of an event handler within the flush?
```

### DIFF
--- a/sqladmin/templates/details.html
+++ b/sqladmin/templates/details.html
@@ -25,8 +25,10 @@
               <a href="{{ model_admin._url_for_details(value) }}">({{ value }})</a>
               {% endfor %}
               </td>
-              {% else %}
+              {% elif model_admin._has_attr(model, attr) %}
               <td><a href="{{ model_admin._url_for_details_with_attr(model, attr) }}">{{ model_admin.get_detail_value(model, attr) }}</a></td>
+              {% else %}
+              <td>{{ model_admin.get_detail_value(model, attr) }}</td>
               {% endif %}
               {% else %}
               <td>{{ model_admin.get_detail_value(model, attr) }}</td>
--- a/sqladmin/templates/details.html
+++ b/sqladmin/templates/details.html
@@ -25,10 +25,8 @@
               <a href="{{ model_admin._url_for_details(value) }}">({{ value }})</a>
               {% endfor %}
               </td>
-              {% elif model_admin._has_attr(model, attr) %}
-              <td><a href="{{ model_admin._url_for_details_with_attr(model, attr) }}">{{ model_admin.get_detail_value(model, attr) }}</a></td>
               {% else %}
-              <td>{{ model_admin.get_detail_value(model, attr) }}</td>
+              <td><a href="{{ model_admin._url_for_details_with_attr(model, attr) }}">{{ model_admin.get_detail_value(model, attr) }}</a></td>
               {% endif %}
               {% else %}
               <td>{{ model_admin.get_detail_value(model, attr) }}</td>
--- a/tests/test_admin_async.py
+++ b/tests/test_admin_async.py
@@ -51,6 +51,7 @@ class User(Base):
     meta_data = Column(JSON)
 
     addresses = relationship("Address", back_populates="user")
+    profile = relationship("Profile", back_populates="user", uselist=False)
 
     def __str__(self) -> str:
         return f"User {self.id}"
@@ -66,6 +67,18 @@ class Address(Base):
 
     def __str__(self) -> str:
         return f"Address {self.id}"
+
+
+class Profile(Base):
+    __tablename__ = "profiles"
+
+    id = Column(Integer, primary_key=True)
+    user_id = Column(Integer, ForeignKey("users.id"), unique=True)
+
+    user = relationship("User", back_populates="profile")
+
+    def __str__(self) -> str:
+        return f"Profile {self.id}"
 
 
 class Movie(Base):
@@ -92,7 +105,14 @@ async def client(prepare_database: Any) -> AsyncGenerator[AsyncClient, None]:
 
 
 class UserAdmin(ModelAdmin, model=User):
-    column_list = [User.id, User.name, User.email, User.addresses, User.status]
+    column_list = [
+        User.id,
+        User.name,
+        User.email,
+        User.addresses,
+        User.profile,
+        User.status,
+    ]
     column_labels = {User.email: "Email"}
     column_searchable_list = [User.name]
     column_sortable_list = [User.id]
@@ -102,6 +122,12 @@ class UserAdmin(ModelAdmin, model=User):
 class AddressAdmin(ModelAdmin, model=Address):
     column_list = ["id", "user_id", "user"]
     name_plural = "Addresses"
+    export_max_rows = 3
+
+
+class ProfileAdmin(ModelAdmin, model=Profile):
+    column_list = ["id", "user_id", "user"]
+    name_plural = "Profiles"
     export_max_rows = 3
 
 
@@ -119,6 +145,7 @@ class MovieAdmin(ModelAdmin, model=Movie):
 
 admin.register_model(UserAdmin)
 admin.register_model(AddressAdmin)
+admin.register_model(ProfileAdmin)
 admin.register_model(MovieAdmin)
 
 
@@ -128,6 +155,7 @@ async def test_root_view(client: AsyncClient) -> None:
     assert response.status_code == 200
     assert response.text.count('<span class="nav-link-title">Users</span>') == 1
     assert response.text.count('<span class="nav-link-title">Addresses</span>') == 1
+    assert response.text.count('<span class="nav-link-title">Profiles</span>') == 1
 
 
 async def test_invalid_list_page(client: AsyncClient) -> None:
@@ -208,6 +236,8 @@ async def test_list_page_permission_actions(client: AsyncClient) -> None:
 
         address = Address(user_id=user.id)
         session.add(address)
+        profile = Profile(user_id=user.id)
+        session.add(profile)
 
     await session.commit()
 
@@ -218,6 +248,13 @@ async def test_list_page_permission_actions(client: AsyncClient) -> None:
     assert response.text.count('<i class="fa-solid fa-trash"></i>') == 10
 
     response = await client.get("/admin/address/list")
+
+    assert response.status_code == 200
+    assert response.text.count('<i class="fa-solid fa-eye"></i>') == 10
+    assert response.text.count('<i class="fa-solid fa-pencil"></i>') == 0
+    assert response.text.count('<i class="fa-solid fa-trash"></i>') == 10
+
+    response = await client.get("/admin/profile/list")
 
     assert response.status_code == 200
     assert response.text.count('<i class="fa-solid fa-eye"></i>') == 10
@@ -245,6 +282,8 @@ async def test_detail_page(client: AsyncClient) -> None:
     for _ in range(2):
         address = Address(user_id=user.id)
         session.add(address)
+    profile = Profile(user_id=user.id)
+    session.add(profile)
     await session.commit()
 
     response = await client.get("/admin/user/details/1")
@@ -260,6 +299,8 @@ async def test_detail_page(client: AsyncClient) -> None:
     assert (
         response.text.count('<a href="/admin/address/details/1">(Address 1)</a>') == 1
     )
+    assert response.text.count("<td>profile</td>") == 1
+    assert response.text.count('<a href="/admin/profile/details/1">Profile 1</a>') == 1
 
     # Action Buttons
     assert response.text.count("http://testserver/admin/user/list") == 2
@@ -338,6 +379,7 @@ async def test_create_endpoint_get_form(client: AsyncClient) -> None:
         '<select class="form-control" id="addresses" multiple name="addresses">'
         in response.text
     )
+    assert '<select class="form-control" id="profile" name="profile">' in response.text
     assert (
         '<input class="form-control" id="name" maxlength="16" name="name"'
         in response.text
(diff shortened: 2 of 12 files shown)


tests/test_admin_sync.py
```diff
@@ -217,6 +247,13 @@ def test_list_page_permission_actions(client: TestClient) -> None:
     assert response.text.count('<i class="fa-solid fa-pencil"></i>') == 0
     assert response.text.count('<i class="fa-solid fa-trash"></i>') == 10
 
+    response = client.get("/admin/profile/list")
+
+    assert response.status_code == 200
+    assert response.text.count('<i class="fa-solid fa-eye"></i>') == 10
+    assert response.text.count('<i class="fa-solid fa-pencil"></i>') == 0
+    assert response.text.count('<i class="fa-solid fa-trash"></i>') == 10
```
aminalaee, line 255: Do we need these tests?
I've been thinking on even reducing these tests in `test_admin_sync.py` and `test_admin_async.py` and at least I could not see the point. For example this is testing the list page which already is tested by two other models. What do you think?

tests/test_admin_sync.py
```diff
@@ -406,6 +479,7 @@ def test_is_visible_method(client: TestClient) -> None:
     assert response.status_code == 200
     assert response.text.count('<span class="nav-link-title">Users</span>') == 1
     assert response.text.count('<span class="nav-link-title">Addresses</span>') == 1
+    assert response.text.count('<span class="nav-link-title">Profiles</span>') == 1
```
aminalaee, line 482: Again I think we can avoid the tests in `test_admin_sync` and `test_admin_async`.

sqladmin/templates/details.html
```diff
@@ -25,8 +25,10 @@ <h3 class="card-title">{{ model_admin.pk_column.name }}: {{ model_admin.get_attr
               <a href="{{ model_admin._url_for_details(value) }}">({{ value }})</a>
               {% endfor %}
               </td>
-              {% else %}
+              {% elif model_admin._has_attr(model, attr) %}
               <td><a href="{{ model_admin._url_for_details_with_attr(model, attr) }}">{{ model_admin.get_detail_value(model, attr) }}</a></td>
+              {% else %}
+              <td>{{ model_admin.get_detail_value(model, attr) }}</td>
```
aminalaee, line 31: Can you please clarify why this is needed?

tests/test_admin_sync.py
okapies: Ok, I'll remove them.

tests/test_admin_sync.py
okapies: :ok_hand:

sqladmin/templates/details.html
okapies: Hmm. I added it because I saw the case `None` is assigned to `model` during writing the tests, but I cannot reproduce it now. :thinking: I'll cancel these changes.

okapies:
I fixed my patch but got an error during running the tests. As described above, I think we need an improvement in handling `session` in the tests. Could you check it?

```python
==================================== ERRORS ====================================
____________________ ERROR at teardown of test_model_setup _____________________
    @pytest.fixture(autouse=True)
    def prepare_database() -> Generator[None, None, None]:
        yield
>       session.close_all()
tests/test_models.py:55: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<string>:2: in close_all
    ???
/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py:401: in warned
    return fn(*args, **kwargs)
/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/sqlalchemy/orm/session.py:81: in close_all
    close_all_sessions()
...
/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/sqlalchemy/dialects/sqlite/aiosqlite.py:202: in rollback
    self.await_(self._connection.rollback())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

awaitable = <coroutine object Connection.rollback at 0x7f8feccf6d50>

     def await_only(awaitable: Coroutine) -> Any:
        ...
        # this is called in the context greenlet while running fn
        current = greenlet.getcurrent()
        if not isinstance(current, _AsyncIoGreenlet):
>           raise exc.MissingGreenlet(
                "greenlet_spawn has not been called; can't call await_only() "
                "here. Was IO attempted in an unexpected place?"
            )
E           sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_only() here. Was IO attempted in an unexpected place? (Background on this error at: https://sqlalche.me/e/14/xd2s)
```

sqladmin/templates/details.html
okapies:
```python
=================================== FAILURES ===================================
______________________________ test_column_labels ______________________________

client = <starlette.testclient.TestClient object at 0x7f22be8b1550>

    def test_column_labels(client: TestClient) -> None:
        user = User(name="Foo")
        session.add(user)
        session.commit()
    
        response = client.get("/admin/user/list")
    
        assert response.status_code == 200
        assert response.text.count("Email") == 1
    
>       response = client.get("/admin/user/details/1")

tests/test_admin_sync.py:310: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.8/site-packages/requests/sessions.py:600: in get
    return self.request("GET", url, **kwargs)
.venv/lib/python3.8/site-packages/starlette/testclient.py:473: in request
    return super().request(
...
.venv/lib/python3.8/site-packages/starlette/templating.py:37: in __init__
    content = template.render(context)
.venv/lib/python3.8/site-packages/jinja2/environment.py:1301: in render
    self.environment.handle_exception()
.venv/lib/python3.8/site-packages/jinja2/environment.py:936: in handle_exception
    raise rewrite_traceback_stack(source=source)
sqladmin/templates/details.html:1: in top-level template code
    {% extends "layout.html" %}
sqladmin/templates/layout.html:1: in top-level template code
    {% extends "base.html" %}
sqladmin/templates/base.html:15: in top-level template code
    {% block body %}
sqladmin/templates/layout.html:49: in block 'body'
    {% block content %} {% endblock %}
sqladmin/templates/details.html:29: in block 'content'
    <td><a href="{{ model_admin._url_for_details_with_attr(model, attr) }}">{{ model_admin.get_detail_value(model, attr) }}</a></td>
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.test_admin_sync.UserAdmin object at 0x7f22bfb20940>
obj = <tests.test_admin_sync.User object at 0x7f22be8e2820>
attr = <RelationshipProperty at 0x7f22bfb72040; profile>

    def _url_for_details_with_attr(self, obj: Any, attr: RelationshipProperty) -> str:
        target = getattr(obj, attr.key)
>       pk = getattr(target, attr.mapper.primary_key[0].name)
E       AttributeError: 'NoneType' object has no attribute 'id'

sqladmin/models.py:592: AttributeError
```

aminalaee: But this code seems outdated to me, this is already fixed and changed: https://github.com/aminalaee/sqladmin/blob/main/sqladmin/models.py#L593

tests/test_forms.py
```diff
@@ -75,6 +93,45 @@ async def prepare_database() -> AsyncGenerator[None, None]:
     await engine.dispose()
 
 
+@pytest.fixture
+async def client(prepare_database: Any) -> AsyncGenerator[AsyncClient, None]:
```
aminalaee, line 97: This is not used.

tests/test_admin_async.py
```diff
@@ -463,6 +533,11 @@ async def test_update_get_page(client: AsyncClient) -> None:
         == 1
     )
     assert response.text.count('<option selected value="1">Address 1</option>') == 1
+    assert (
+        response.text.count('<select class="form-control" id="profile" name="profile">')
```
aminalaee, line 537: Sorry I think I wasn't clear in my previous comment, I think we should get rid of adding tests in both `test_admin_sync` and `test_admin_async` unless we have a very good reason.
For example this is checking the `select` which is already tested by OneToMany relation. I think we can avoid both these files changes.

aminalaee:
> I fixed my patch but got an error during running the tests. As described above, I think we need an improvement in handling session in the tests. Could you check it?

Yes I think if we revert the tests I mentioned it will go away, but I will work on improving those tests in a separate PR.

tests/test_admin_async.py
okapies: Thanks for the follow-up. I'm sorry but I can't judge which tests are really unnecessary because I don't fully understand your implementations. I think there are many potential pitfalls between `OneToOne` (single object) and `OneToMany` (array of objects), so I basically added the checks same as for `Address`.

For example, the above test checks that the response has a `<select>` tag *without* `multiple` attribute. IMO, these tests are verbose but it seems to be useful to detect some kind of degradation in the future.

So, I want you to merge it as it is, and remove ones you think unnecessary in another patch (or remove all tests to add them later?). What do you think?

aminalaee: Yeah that's ok, I can clean it up which I will need to do anyway, let's then focus to make this working and merge it as is.

okapies: Thanks!

aminalaee (approved):
Thank you @okapies very much for the contribution.

okapies:
🙇

okapies:
Hi @aminalaee, could you please release the current `master` including this and #167? I'd like to use it in my project.

aminalaee:
@okapies Hi Yes sure, that's my plan after two pending PRs are merged today.

okapies:
I see. Thanks so much!

PR #184, opened 2022-06-17
Fix form fields order when specifying columns

### LINKED ISSUE #181: Show the form fields in the order of `only`
### Checklist

`get_model_form()` receives `only` parameter in `Sequence` but ignores its order to build form fields.

`attributes` should align to the order of `only` if specified like:

```python
    attributes = []
    names = only or mapper.attrs.keys()
    for name in names:
        if exclude and name in exclude:
            continue
        attributes.append((name, mapper.attrs[name]))
```

_No response_

### Additional context

_No response_

okapies:
Fixes #181.

### DIFF
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -400,13 +400,11 @@ async def get_model_form(
     form_overrides = form_overrides or {}
 
     attributes = []
-    for name, attr in mapper.attrs.items():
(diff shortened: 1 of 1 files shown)


aminalaee (approved):
Thanks a lot for the contribution 👍

PR #186, opened 2022-06-19
Fix ModelConverter when impl is not callable

aminalaee:
Supersedes #177


aminalaee:
Thanks to @okapies for raising the issue and creating the initial PR.

PR #204, opened 2022-06-22
Fix handling of iterable fields

okapies:
Fixes #196.
Fixes #203.

PR #206, opened 2022-06-22
Use zip for iterating over the formatted values

okapies:
Fix to use `zip` for iterating over the formatted values. It prevents retrieving `formatted_value` by index especially when it is `Iterable` but not `Sequence`.

Ref. #204.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -52,6 +52,7 @@ class BaseAdmin:
             ]
         )
         self.templates.env.globals["min"] = min
+        self.templates.env.globals["zip"] = zip
         self.templates.env.globals["admin_title"] = title
         self.templates.env.globals["admin_logo_url"] = logo_url
         self.templates.env.globals["model_admins"] = self.model_admins
--- a/sqladmin/templates/details.html
+++ b/sqladmin/templates/details.html
@@ -22,8 +22,8 @@
               {% if (name, attr) in model_admin._details_relations %}
               {% if is_iterable( value ) %}
               <td>
-              {% for elem in value %}
-              <a href="{{ model_admin._url_for_details(elem) }}">({{ formatted_value[loop.index0] }})</a>
+              {% for elem, formatted_elem in zip(value, formatted_value) %}
+              <a href="{{ model_admin._url_for_details(elem) }}">({{ formatted_elem }})</a>
               {% endfor %}
               </td>
               {% else %}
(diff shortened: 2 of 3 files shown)


aminalaee:
I think we could also update `is_iterable` to `isinstance(x, Sequence) or isinstance(x, InstrumentedList)`.
The main reason we switched from `Sequence` to `Iterable` was that SQLAlchemy `InstrumentedList` was not a Sequence.

okapies:
[By definition](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), `Sequence` is a trait both of `Reversible` and `Collection` with `__getitem__`, `index` and `count`. In this sense, `InstrumentedList` is actually a `Sequence` even if it is not a subclass of it. And [thanks to `abc` module](https://docs.python.org/3/library/collections.abc.html#id17), you don't need to check if the value is `InstrumentedList`:

```python
>>> from collections.abc import Sequence
>>> from sqlalchemy.orm.collections import InstrumentedList
>>> isinstance(InstrumentedList(), Sequence)
True
```

But, using the minimum interface for treating value is generally good. If we just need to iterate over the value to render, we should use `Iterable` not `Sequence`. If we plan to use `__getitem__`, `__contains__` and `__len__` in the future, we should choose `Sequence`. It depends on the requirement for the design of this library.

okapies:
Anyway, I think merging this PR will not affect your design decision because `Iterable` is less strict than `Sequence`.

aminalaee:
@okapies Yes you're right about the definitions, my comment was not really accurate.
Previous instead of `is_iterable` it was `is_list` and we were checking `isinstance(x, list)` but because there was an issue distinguishing SQLAlchemy `InstrumentedList` we exactly switched to `abc` `Sequence` and `Iterable`.

But since `isinstance(InstrumentedList(), list) == True` then I'm not sure why it should be `is_iterable` when it can be `is_list` simply. Let me dig a bit into the older PRs.

aminalaee:
I know this won't discard your PR, using `zip` would still make it more readable in template, it will just make the type checking of `is_iterable` or `is_list` strict.

okapies:
Ok. Anyway, could you please release a new version including #204? It is a show-stopper for our app.

aminalaee:
> Ok. Anyway, could you please release a new version including #204? It is a show-stopper for our app.

Yes sure, I'll merge these and do the release.

aminalaee (approved):
Thanks for the PR 👍

okapies:
:bow:

PR #207, opened 2022-06-22
Add form_include_pk option

### LINKED ISSUE #200: Don't hide primary keys in creation forms
### Checklist

Currently, all primary (and foreign) key columns are discarded from models when creating forms.

This presents a problem with models where primary keys are not auto-generated.
For example:
```python
class User(Base):
    name = Column(String, primary_key=True)
    age = Column(Integer)
```
`User` creation form only displays the input field for `age`, not for `name`; and the creation fails because `name` isn't nullable.

### Solution 1
Add ability to configure this behavior when subclassing `ModelAdmin`.
Perhaps a `form_exclude_primary_keys` attribute.

### Solution 2
Always show primary keys in forms and sort out nullability.

_No response_

### Additional context

Columns get discarded here:
https://github.com/aminalaee/sqladmin/blob/43f1fc1c218fd68be5b410edb16fa7288f498b44/sqladmin/forms.py#L117-L118

aminalaee:
Fixes #200 gives the option to include PK in create/edit forms.

Before you had to do this:

```python
class UserAdmin(ModelAdmin, model=User):
    form_columns = [User.id, ....]
```

Now if you don't want to edit form_columns you can do:

```python
class UserAdmin(ModelAdmin, model=User):
    form_include_pk = True
```

### DIFF
--- a/docs/api_reference/model_admin.md
+++ b/docs/api_reference/model_admin.md
@@ -40,3 +40,4 @@
         - form_excluded_columns
         - form_overrides
         - form_widget_args
+        - form_include_pk
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -179,6 +179,7 @@ The forms are based on `WTForms` package and include the following options:
 * `form_columns`: List of model columns to be included in the form. Default is all model columns.
 * `form_excluded_columns`: List of model columns to be excluded from the form.
 * `form_overrides`: Dictionary of form fields to override when creating the form.
+* `form_include_pk`: Control if primary key column should be included in create/edit forms. Default is `False`.
 
 !!! example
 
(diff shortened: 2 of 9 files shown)


afparsons:
@aminalaee 

This is extremely confusing behavior!

```python
if (column.primary_key or column.foreign_keys) and not form_include_pk:
    return None
```

specifically this part:

```python
or column.foreign_keys
```

From the documentation, I expect this feature to work like this:

```python
if column.primary_key and not form_include_pk:
    return None
```

I'm glad I found this relatively quickly because I was very confused as to why foreign keys disappeared from the model form.

Maybe better to split this into `form_include_pk` and `form_include_fk` ?

aminalaee:
@afparsons Why is that confusing? What is your exact use-case?
You wouldn't typically need the FK columns as the relationship properties would be used.

afparsons:
@aminalaee, I could very well be making a mistake here; I will try to come up with a trivial example in the next few days and post.

(I'm also re-reading my above comment. Just for clarity, it wasn't supposed to be too negative. I'm very grateful for this project and I am extremely impressed with how responsive you are!)

PR #211, opened 2022-06-23
Fix lazy subquery in list query

lebrunthibault:
See https://github.com/aminalaee/sqladmin/issues/210


lebrunthibault:
Handled by https://github.com/aminalaee/sqladmin/pull/212

PR #214, opened 2022-06-25
Add time field converter

ischaojie:
Looks like there is no time field converter :)

### DIFF
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -12,6 +12,7 @@ from sqlalchemy import (
     Numeric,
     String,
     Text,
+    Time,
     TypeDecorator,
 )
 from sqlalchemy.dialects.postgresql import INET, MACADDR, UUID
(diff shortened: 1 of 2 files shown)


ischaojie:
Fix the lint :)

ischaojie:
I'm not sure what this means, `cadecov/patch` failed.

aminalaee:
There's no test for this, it's missing coverage. There's a model which has all the fields for coverage, you can add it there.

ischaojie:
> There's no test for this, it's missing coverage. There's a model which has all the fields for coverage, you can add it there.

Got it!

ischaojie:
Add a test for this:)

tests/test_forms.py
```diff
@@ -243,3 +244,14 @@ class CustomModel(Base):
 async def test_model_form_include_pk() -> None:
     Form = await get_model_form(model=User, engine=engine, form_include_pk=True)
     assert "id" in Form()._fields
+
+
+async def test_form_time_converter() -> None:
+    class ExampleModel(Base):
+        __tablename__ = "example_model"
+        id = Column(Integer, primary_key=True)
+        start_time = Column(Time)
```
aminalaee, line 253: You can add this new field to here, I think it won't need a separate test.

aminalaee: I mean if you add this Time field to the `User` class in there, you can just assert it in `test_model_form` and no need for the new test.

ischaojie: got it.

PR #220, opened 2022-06-29
Adds Pre-commit

colin99d:
Adds a setup file for pre-commit. This allows the linters to be ran in local before pushing a pr. To install do the following:

- Type in `pip install pre-commit`
- Inside an individual repo type in `pre-commit install`
- After this users will have to fix linting issues before a commit will go through.

Note: The pre-commit hooks can be skipped by adding `-n` to a commit

The following linters were added:

- autoflake: used in github actions
- isort: used in github actions
- black: used in github actions
- flake8: used in github actions
- Some basic builtins: check-yaml, end-of-file-fixer, trailing-whitespaces, check-merge-conflict, detect-private-key
- codespell: makes sure that words in code are spelled correctly
- pyupgrade: checks for outdated syntax in python


aminalaee:
This looks good, thanks.
My only concern is that we already have these in `scripts/check` bash script, can we somehow make use of that instead of re-defining these again?

colin99d:
Also im not sure if the output would look as good.
Screen Shot 2022-06-29 at 9 26 18 AM

aminalaee:
Yeah I also use `check` on my local, so I was thinking if we could use them as git hooks.

colin99d:
I will look into this this weekend.

aminalaee:
Yeah I think if we run `scripts/check` in pre-commit that should be alright.

colin99d:
Much simpler!

aminalaee (approved):
Thank you for the contribution 🚀

PR #228, opened 2022-07-05
Add RadioField custom model column

ischaojie:


sqladmin/widgets.py
```diff
@@ -68,3 +69,62 @@ def __call__(self, field: Field, **kwargs: Any) -> str:
         kwargs.setdefault("data-role", "timepicker")
         kwargs.setdefault("data-date-format", "HH:mm:ss")
         return super().__call__(field, **kwargs)
+
+
+class ListWidget:
```
aminalaee, line 74: Can't we use the `ListWidget from the WTForms package?

sqladmin/fields.py
```diff
@@ -419,3 +421,25 @@ def pre_validate(self, form: Form) -> None:
                 identity = inspect(v).identity
                 if identity and str(identity[0]) not in pk_list:  # pragma: no cover
                     raise ValidationError(self.gettext("Not a valid choice"))
+
+
+class RadioField(SelectField):
```
aminalaee, line 426: And also WTForms already has a RadioField.

sqladmin/widgets.py
ischaojie: The reason is default `ListWidget` is not suited for tabler radio style.

aminalaee: Can you explain please what is the difference? Do we have to create another one? Can't we inherit and modify it?

ischaojie: The default `ListWidget` use html ul or ol tag, but in tabler it looks ugly:

image

For your advice, maybe can inherit from it.

ischaojie: But wtforms ListWidget force tag in `<ol>` and `<ul>` and subfield tag in `<li>`:

https://github.com/wtforms/wtforms/blob/master/src/wtforms/widgets/core.py#L100

ischaojie: Or we can add a custom style to change this.

ischaojie:
It's been too long for this pr, close it first.

PR #229, opened 2022-07-05
Fix missing browser tab title

cuamckuu:
`<title>` tag wasn't set propperly in `<head>` of html, so I fixed it


aminalaee (approved):
Great, thank you.

PR #232, opened 2022-07-07
Adds better example column_formatter callback docs

jordotech:
Example renders booleans as icons


aminalaee:
Interesting..I actually had a note to enable this by default and allow configurations to toggle this. I think most people would like to see this by default. That's the same in Flask-Admin.

aminalaee:
If you're interested please update this so instead this is the default behaviour in `SQLAdmin`.

Issue #237, opened 2022-07-12
Custom page creation


aminalaee:
### Discussed in https://github.com/aminalaee/sqladmin/discussions/236

<div type='discussions-op-text'>

<sup>Originally posted by **masksshow** July 12, 2022</sup>
Hello.
Can I add a custom page to the menu? It is not related to the model.
For example, a statistics page or a custom template.</div>

just-an-dev:
I also looking for that, and I look a bit how to do it. 
@aminalaee Do you have idea of "best way" to do that ?
Adding some parameters/method to BaseAdmin to only add it in menu the "route" can be a good start maybe ?

just-an-dev:
I have something working, but I need to look how refactoring to be clean. And remove FastAPI dependency.

aminalaee:
Cool, feel free to create a PR for it.

just-an-dev:
I just open one to have "feedback" and talk on how to do it in clean way, because it's very ugly 😂

aminalaee:
Thanks to @just-an-dev This is merged and will be in `0.2.0` release

PR #244, opened 2022-07-20
Add custom page

just-an-dev:
Goal: is to be able to have a custom page, linked to #237

- [x] Remove FastAPI dependency
- [x]  Menu management to review
- [x] Remove dirty sample/tests
- [x] Update documentation

# Question

Menu management : Add a class over ModelAdmin, to only handle menu related data (url, icon, name, ....) ?

Use a class view for only one endpoint/page ? or one class an handle multiple page/route ?


sqladmin/models.py
```diff
@@ -125,7 +126,24 @@ def is_accessible(self, request: Request) -> bool:
...
+
+    name: ClassVar[str]
+    icon: ClassVar[str]
+    path: ClassVar[str]
+    methods = ["GET"]
+    endpoint: ClassVar[Callable]
+    include_in_schema: ClassVar[bool] = True
+
+    def render(self, template, data):
+        return self.templates.TemplateResponse(template, data)
```
just-an-dev, line 143: Maybe this function is useless, let the user directly call the `self.templates.TemplateResponse(...)` ?

aminalaee: yeah I agree it will be simpler without this.

sqladmin/application.py
```diff
@@ -103,6 +117,27 @@ class UserAdmin(ModelAdmin, model=User):
...
 
+    def register_view(self, view: Type["ModelView"]):
+
+        class_view = view()
+        class_view.name_plural = class_view.name
+        class_view.url_path_for = self.app.url_path_for
+
+        # add template directory
+        class_view.templates = self.init_templating_engine(
+            extra_loader=FileSystemLoader(class_view.template_path)
```
aminalaee, line 128: Maybe it's more useful to accept one `templates_path` with `Admin(...)` instead of setting it per view?

aminalaee:
Thank you for the PR, I'll review it over the weekend.

sqladmin/application.py
```diff
@@ -103,6 +113,22 @@ class UserAdmin(ModelAdmin, model=User):
 
         self._model_admins.append((model()))
 
+    def register_view(self, view: Type["ModelView"]) -> None:
+        class_view = view()
+        class_view.name_plural = class_view.name
+        class_view.url_path_for = self.app.url_path_for
+        class_view.templates = self.templates
```
just-an-dev, line 120: May the `self.templates` can be also added into `register_model` ?

I'm also working on PR to be able to add custom action in the list page for models, and it can be useful :)

example.py
```diff
@@ -0,0 +1,49 @@
...
+
+class CustomAdmin(ModelView):
+    def test_html(self, request: Request):
+        return self.templates.TemplateResponse("sample.html", context={"request": request})
+
+    name = "Test me"
+    icon = "fa-user"
+    path = "/custom/hmlt"
+    methods = ["GET"]
+    endpoint = test_html
```
just-an-dev, line 46: Need a better way to define routing settings !

sqladmin/models.py
```diff
@@ -125,7 +125,18 @@ def is_accessible(self, request: Request) -> bool:
         return True
 
 
-class ModelAdmin(BaseModelAdmin, metaclass=ModelAdminMeta):
+class ModelView(BaseModelAdmin):
```
aminalaee, line 128: Maybe this can be something like `View` or `BaseView`?
I think the name `Model...` is reflecting database model and it's not entirely correct in this case.

sqladmin/application.py
```diff
@@ -170,6 +196,7 @@ def __init__(
         logo_url: str = None,
         middlewares: Sequence[Middleware] = None,
         debug: bool = False,
+        template_path: str = None,
```
aminalaee, line 199:
```suggestion
        templates_dir: str = "templates",
```

sqladmin/application.py
```diff
@@ -103,6 +113,22 @@ class UserAdmin(ModelAdmin, model=User):
 
         self._model_admins.append((model()))
 
+    def register_view(self, view: Type["ModelView"]) -> None:
+        class_view = view()
```
aminalaee, line 117: I think this is the other way around, the class_view should be the argument because it is Type[...] and `view` should be instantiated from it, what do you think?

sqladmin/application.py
aminalaee: Thanks for the suggestion, but I think if it's not required here let's keep it in that separate PR.

sqladmin/application.py
```diff
@@ -36,28 +35,39 @@ def __init__(
         base_url: str = "/admin",
         title: str = "Admin",
         logo_url: str = None,
+        template_path: str = None,
```
aminalaee, line 38: Same default value here.

sqladmin/application.py
```diff
@@ -36,28 +35,39 @@ def __init__(
...
-            ]
-        )
-        self.templates.env.globals["min"] = min
-        self.templates.env.globals["zip"] = zip
-        self.templates.env.globals["admin_title"] = title
-        self.templates.env.globals["admin_logo_url"] = logo_url
-        self.templates.env.globals["model_admins"] = self.model_admins
-        self.templates.env.globals["is_list"] = lambda x: isinstance(x, list)
+        self.title = title
+        self.logo_url = logo_url
```
aminalaee, line 44: Do we need to assign these to the class? If it's only the usage in `init_template_engine` can we just pass them directly?

sqladmin/application.py
```diff
@@ -36,28 +35,39 @@ def __init__(
...
+        self.template_path = template_path
+        self._model_admins: List["ModelView"] = []
+
+        self.templates = self.init_templating_engine()
+
+    def init_templating_engine(self) -> Jinja2Templates:
+        templates = Jinja2Templates("templates")
+        loaders = [
+            FileSystemLoader("templates"),
```
aminalaee, line 53:
```suggestion
            FileSystemLoader(self.templates_dir),
```

Should be ok. Since we can default it to `templates`.

aminalaee (changes requested):
Thanks again for the changes 👍

sqladmin/models.py
just-an-dev: yeah BaseView is more clear :)

sqladmin/application.py
just-an-dev: yes no really need to add this to class

sqladmin/application.py
```diff
@@ -13,9 +13,7 @@
 from starlette.staticfiles import StaticFiles
 from starlette.templating import Jinja2Templates
 
-if TYPE_CHECKING:
-    from sqladmin.models import ModelAdmin
```
just-an-dev, line 17: I have remove this check to be able to pass the test script successfully, and don't get any circular import errors, did you remember the case need this check ?

aminalaee: Not really, how can it create circular import, it's weird, let me see.

just-an-dev:
i looked why the code coverage decrease, and it seems `tests` directory is also take in account, maybe we need to exclude it to avoid checking coverage of tests

tests/test_view.py
```diff
@@ -0,0 +1,57 @@
...
+    path = "/custom/test_page"
+    methods = ["GET"]
+    endpoint = test_page
+
+
+@pytest.fixture(autouse=True)
+def prepare_database() -> Generator[None, None, None]:
+    Base.metadata.create_all(engine)
+    yield
+    Base.metadata.drop_all(engine)
```
aminalaee, line 49: I'm not sure why coverage is decreased, but there are a bunch of things which are not needed (I'm guessing this fixture could be also the reason, not sure).

You can remove `User` and  database related things and this fixture

tests/test_view.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_register_view() -> None:
+    admin = Admin(app=Starlette(), engine=engine)
+    admin.register_view(CustomAdmin)
+
+    url = CustomAdmin().url_path_for(CustomAdmin.name_plural)
+    assert url == "/custom/test_page"
```
aminalaee, line 57: I'd add the test to see if it's rendering the template correctly.
You can create a `templates` dir in the tests and add your HTML in there for the test.

example/app.py
```diff
@@ -0,0 +1,49 @@
...
+        return self.templates.TemplateResponse("custom.html", context={"request": request})
+
+    name_plural = "Test me"
+    icon = "fa-user"
+    path = "/custom/test_page"
+    methods = ["GET"]
+    endpoint = test_page
+
+
+admin.register_view(CustomAdmin)
```
aminalaee, line 49: And the example is not needed right? We can have the docs and docstrings in place for these configurations.

sqladmin/application.py
```diff
@@ -115,8 +121,51 @@ class UserAdmin(ModelAdmin, model=User):
...
+        class_view = view()
+        self._model_admins.append(class_view)
+
+        self.app.add_route(
+            route=class_view.endpoint,
+            path=class_view.path,
+            methods=class_view.methods,
+            name=class_view.name_plural,
+            include_in_schema=class_view.include_in_schema,
+        )
```
aminalaee, line 161:
```suggestion
        self._model_admins.append(view())

        self.app.add_route(
            route=view.endpoint,
            path=view.path,
            methods=view.methods,
            name=view.name_plural,
            include_in_schema=view.include_in_schema,
        )
```

aminalaee (approved):
Thanks a lot for the PR. this is a great starting point, I'll update the API a bit later.

PR #251, opened 2022-07-31
Add `expose` for BaseView

aminalaee:
Adds `expose` method to be used for `BaseView` like:

```python
class CustomAdmin(BaseView):
    name = "test"
    icon = "fa fa-test"

    @expose("/custom", methods=["GET"])
    def custom(self, request: Request):
        return self.templates.TemplateResponse(
            "custom.html", context={"request": request}
        )

    @expose("/custom/report")
    def custom_report(self, request: Request):
        return self.templates.TemplateResponse(
            "custom.html", context={"request": request}
        )
```


aminalaee:
@just-an-dev if you'd like to take a look.

PR #273, opened 2022-08-08
Add possibility to add some actions on models

just-an-dev:
Add possibility to create custom method to interact with the models, and have an button to trigger from interface

For example you can add a method to trigger a specific logic on an user :
 
```python
app = FastAPI()
admin = Admin(app, engine)


    @action(name='Trigger Update', add_in_detail=True, add_in_list=True)
    async def action_on_user(self, request: Request):
        model = await self.get_model_by_pk(request.path_params["pk"])

        return self.templates.TemplateResponse("sample.html", {"request": request, "data": f'user : {model.id}'})


admin.add_view(UserAdmin)

```
 
image

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -275,3 +275,27 @@ from sqladmin import Admin
 
 admin = Admin(templates_dir="my_templates", ...)
 ```
+
+
+## Custom Action
+
+To add custom action on models to the Admin interface, you can use the `action` annotation.
+
+For example:
+
+!!! example
+
+    ```python
+    from sqladmin import BaseView, action
+
+    class UserAdmin(ModelView, model=User):
+    
+        @action(name='Trigger Update', add_in_detail=True, add_in_list=True)
+        async def action_on_user(self, request: Request):
+            model = await self.get_model_by_pk(request.path_params["pk"])
+    
+            ...
+    
+    admin.add_view(UserAdmin)
+    ```
+
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -290,11 +290,17 @@ For example:
 
     class UserAdmin(ModelView, model=User):
     
-        @action(name='Trigger Update', add_in_detail=True, add_in_list=True)
+        @action(name='approve_user',label="Approve", confirmation_message="Are you sure ?", add_in_detail=True, add_in_list=True)
         async def action_on_user(self, request: Request):
-            model = await self.get_model_by_pk(request.path_params["pk"])
+            model: User = await self.get_model_by_pk(request.path_params["pk"])
+
+            with self.sessionmaker(expire_on_commit=False) as session:
+                ...
+                session.add(model)
+                session.commit()
     
-            ...
+            detail_url = self._url_for_details(model)
+            return Response(content=None)
     
     admin.add_view(UserAdmin)
     ```
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -20,6 +20,7 @@ from sqladmin.models import BaseView, ModelView
 __all__ = [
     "Admin",
     "expose",
+    "action",
 ]
 
 
@@ -131,7 +132,26 @@ class BaseAdmin:
             )
             view.async_engine = True
 
-        self._views.append((view()))
+        view_instance = view()
+        funcs = inspect.getmembers(view_instance, predicate=inspect.ismethod)
+
+        for _, func in funcs[::-1]:
+            if hasattr(func, "_action"):
+                self.admin.add_route(
+                    route=func,
+                    path="/{identity}/action/{pk}/{action_name}",
+                    methods=["GET", "POST"],
+                    name=f"{view_instance.identity}-{func._name}",
+                    include_in_schema=func._include_in_schema,
+                )
+
+                if func._add_in_list:
+                    view.custom_actions_in_list.append(func._name)
+                if func._add_in_detail:
+                    view.custom_actions_in_detail.append(func._name)
+
+        view.templates = self.templates
+        self._views.append((view_instance))
 
     def add_base_view(self, view: Type[BaseView]) -> None:
         """Add BaseView to the Admin.
@@ -479,7 +499,7 @@ def expose(
     *,
     methods: List[str] = ["GET"],
     identity: str = None,
-    include_in_schema: bool = True
+    include_in_schema: bool = True,
 ) -> Callable[..., Any]:
     """Expose View with information."""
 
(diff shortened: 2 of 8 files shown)


aminalaee:
Hey thanks a lot for the PR, this is a great idea.
Just need to first see if this exists in Flask-Admin then do it similar to Flask-Admin, and if not, review the API and feature 👍

just-an-dev:
I have see something like that on Flask-Admin here : https://flask-admin.readthedocs.io/en/latest/advanced/?highlight=action#customizing-batch-actions

just-an-dev:
I have made something more to add custom page for a model, but maybe it will be better to stick to Flask-Admin and use the `@action` annotation to make more backend action, instead of displaying a page ? ( and adding the confirmation box, etc, ...)

aminalaee:
Yeah I think for a start wr can do exactly as Flask-Admin and then build on top of that.

just-an-dev:
i finally got something working and have same feature of Flask-Admin, added a little confirmation modal for actions

sqladmin/application.py
```diff
@@ -493,3 +518,28 @@ def wrap(func):
         return func
 
     return wrap
+
+
+def action(
```
aminalaee, line 523: Maybe this should go to `sqladmin.actions`.

sqladmin/application.py
```diff
@@ -131,7 +132,31 @@ class UserAdmin(ModelView, model=User):
             )
             view.async_engine = True
 
-        self._views.append((view()))
+        view_instance = view()
+        funcs = inspect.getmembers(view_instance, predicate=inspect.ismethod)
```
aminalaee, line 136: This is doing something very similar to `expose` we should probably see if there's a chance to to re-use any of that, I'll take a look too.

sqladmin/models.py
```diff
@@ -448,6 +448,18 @@ class UserAdmin(ModelView, model=User):
     Unlimited by default.
     """
 
+    custom_actions_in_list: ClassVar[Dict[str, str]] = {}
```
aminalaee, line 451: If I understand correctly these are only for internal configurations, not for setup by user.
In that case you can just set them up in `__init__` with `self._ custom_actions_in_list ` to mark internal.

sqladmin/models.py
```diff
@@ -712,6 +724,14 @@ def _url_for_details_with_attr(self, obj: Any, attr: RelationshipProperty) -> st
             pk=pk,
         )
 
+    def _url_for_action(self, obj: Any, action_name: str) -> str:
+        pk = getattr(obj, inspect(obj).mapper.primary_key[0].name)
```
aminalaee, line 728: I think the object passed here is an instance of the current ModelView, so maybe this should work:
```python
pk = getattr(obj, self.pk)
```

tests/test_models_action.py
```diff
@@ -0,0 +1,100 @@
...
+        add_in_detail=False,
+        add_in_list=True,
+    )
+    async def send_notification_user(self, request: Request):
+        model = await self.get_model_by_pk(request.path_params["pk"])
+
+        detail_url = self._url_for_details(model)
+        return Response(
+            content=detail_url
+        )  # redirect to a specific url / use None to return to same page
```
aminalaee, line 50: Can you also explain this?

aminalaee (commented):
This looks great, thanks, maybe we need to review it again but this is a very useful feature.

sqladmin/templates/list.html
```diff
@@ -121,6 +124,26 @@ <h3 class="card-title">{{ model_view.name_plural }}</h3>
...
+                        <a href="#" id="link-action-confirm-{{ custom_action }}" data-name="{{ custom_action }}"
+                        data-url="{{ model_view._url_for_action(row, custom_action) }}" data-bs-toggle="modal"
+                        data-bs-target="#modal-confirmation-{{ custom_action }}" class="btn btn-primary">
+                        {{ label }}
+                        </a>
+                        {% with confirmation_message = model_view.custom_actions_confirmation[custom_action], custom_action=custom_action %}
+                        {% include 'modals/confirmation.html' %}
+                        {% endwith %}
+                    {% else %}
+                        <a href="{{ model_view._url_for_action(row, custom_action) }}" class="btn btn-primary">
```
aminalaee, line 140: Is this for handling the logic if no confirmation is set, then no modal is shown?

tests/test_models_action.py
just-an-dev: I return an specific url to redirect user to specific page if needed with the javascript, like on delete modal, with the `window.location.href = result;`

sqladmin/models.py
just-an-dev: yeah it make sense to move it into ModelView as private

sqladmin/application.py
just-an-dev: you mean move it to new file ?

sqladmin/application.py
just-an-dev: Yeah maybe we can refactor it, maybe after this one reviewed/merged

sqladmin/templates/list.html
just-an-dev: yeah only a link is showed, and user is free to redirect to a page/show what they want. But maybe not required, or add a param in `@action` ... like you feel is better :)

sqladmin/models.py
just-an-dev: or maybe `getattr(obj, get_primary_key(obj).name)` like in other `_url_for_delete/_url_for_edit/_url_for_details` ?

aminalaee: My point was to avoid calling `inspect` again, but `get_primary_key()` will do that.
I think my comment should've been `pk = getattr(obj, self.pk.name)`, doesn't this work?

just-an-dev: hoo ok, i don't see any `pk` attribute ... or maybe i'm not looking the good object

just-an-dev: my bad it's `getattr(obj, self.pk_column.name)` not `pk`

docs/configurations.md
```diff
@@ -275,3 +275,33 @@ from sqladmin import Admin
...
+For example:
+
+!!! example
+
+    ```python
+    from sqladmin import BaseView, action
+
+    class UserAdmin(ModelView, model=User):
+    
+        @action(name='approve_user',label="Approve", confirmation_message="Are you sure ?", add_in_detail=True, add_in_list=True)
```
aminalaee, line 293: Thanks again for the PR,
After taking a look into what Flask-Admin does I have some ideas:

- In the list page, we should be able to select multiple records and apply apply this action to batch of items, so I think both UI and Backend should change for that.
- I think it'd be easier to review if we separate detail and list PRs
- I think we should either some modal, or call that URL with Ajax, I think the redirect is not much useful and redirecting back from that URL to Admin can be hard.
- I think we should update `actions` to re-use existing code before merging.

aminalaee (changes requested):

aminalaee:
Are you interested to address the comments and resolve conflicts?

PR #276, opened 2022-08-22
Add model form converter

shvydkyid:


aminalaee:
Thanks for the PR,
Can you please update the description? And also explain any specific reason you need this? The out-of-the-box conversion does not work in your case?

PR #281, opened 2022-08-25
Switch to hatchling commands and remove scripts

aminalaee:
Switch to Hatchling for package management.

Remove scripts directory and use Hatch commands.


ofek:
Awesome use of scripts!!!

aminalaee:
@ofek Thanks for the great work :)

pyproject.toml
```diff
@@ -50,3 +51,59 @@ include = [
 exclude = [
   "tests/*",
 ]
+
+[tool.hatch.envs.default]
+dependencies = [
```
Kludex, line 56: Does dependabot work on those?

ofek: No

Kludex: this project uses dependabot, so I guess it should replace by https://github.com/renovatebot/renovate ?

ofek: seems stuck in the past too https://docs.renovatebot.com/configuration-options/#python

Issue #304, opened 2022-09-06
Error when sorting with the label


GitBib:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

When a field is sorted and has a specified label, the following error occurs: `sqlalchemy.exc.CompileError: Can't resolve label reference for ORDER BY / GROUP BY / DISTINCT etc. Textual SQL expression 'Company type' should be explicitly declared as text('Company type')`.

### Environment

Python3.10, sqladmin 0.4.0, PostgreSQL 14.1

aminalaee:
Closed by #305

PR #305, opened 2022-09-06
Fix Lebel in sorting

GitBib:
This fixes the problem #304


aminalaee (approved):
Perfect, thanks for the contribution.
Just in time for the release 👍

PR #336, opened 2022-09-24
Fix Boolean field

aminalaee:
Fixes https://github.com/aminalaee/sqladmin/issues/330
A regression in https://github.com/aminalaee/sqladmin/pull/321

Needs to change:
- Boolean field not nullable -> Checkbox
- Boolean field nullable -> SelectField with `True`, `False` and `None`


sqladmin/forms.py
```diff
@@ -337,7 +337,15 @@ def conv_text(
...
-        return BooleanField(**kwargs)
+        if not prop.columns[0].nullable:
+            kwargs.setdefault("render_kw", {})
+            kwargs["render_kw"]["class"] = "form-check-input"
+            return BooleanField(**kwargs)
+
+        kwargs["allow_blank"] = True
+        kwargs["choices"] = [(True, "True"), (False, "False")]
+        kwargs["coerce"] = lambda v: v == "True"
+        return SelectField(**kwargs)
```
aminalaee, line 348: It's strange that selecting `None` and `False` from UI work fine and as expected.
But selecting `True` gives `Not a valid choice`.

PR #342, opened 2022-09-26
Add on_model_change and after_model_change methods

dima23113:
Hello, this is my first pull request. Added a method for editing data before saving it to the database. Flask admin analog

### DIFF
--- a/sqladmin/_queries.py
+++ b/sqladmin/_queries.py
@@ -109,8 +109,10 @@ class Query:
 
         with self.model_view.sessionmaker() as session:
             obj = session.execute(stmt).scalars().first()
+            anyio.from_thread.run(self.model_view.on_model_change, data, obj, False)
             obj = self._set_attributes_sync(session, obj, data)
             session.commit()
+            anyio.from_thread.run(self.model_view.after_model_change, data, obj, False)
 
     async def _update_async(self, pk: Any, data: Dict[str, Any]) -> None:
         pk = get_column_python_type(self.model_view.pk_column)(pk)
@@ -122,8 +124,10 @@ class Query:
         async with self.model_view.sessionmaker() as session:
             result = await session.execute(stmt)
             obj = result.scalars().first()
+            await self.model_view.on_model_change(data, obj, False)
             obj = await self._set_attributes_async(session, obj, data)
             await session.commit()
+            await self.model_view.after_model_change(data, obj, False)
 
     def _delete_sync(self, obj: Any) -> None:
         with self.model_view.sessionmaker() as session:
--- a/sqladmin/_queries.py
+++ b/sqladmin/_queries.py
@@ -143,17 +143,21 @@ class Query:
         obj = self.model_view.model()
 
         with self.model_view.sessionmaker() as session:
+            anyio.from_thread.run(self.model_view.on_model_change, data, obj, True)
             obj = self._set_attributes_sync(session, obj, data)
             session.add(obj)
             session.commit()
+            anyio.from_thread.run(self.model_view.after_model_change, data, obj, True)
 
     async def _insert_async(self, data: Dict[str, Any]) -> None:
         obj = self.model_view.model()
 
         async with self.model_view.sessionmaker() as session:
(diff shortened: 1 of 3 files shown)


sqladmin/models.py
```diff
@@ -942,6 +942,12 @@ def get_export_columns(self) -> List[Tuple[str, MODEL_ATTR_TYPE]]:
             default=[item[1] for item in self._list_attrs],
         )
 
+    def on_model_change(self, data: dict, pk: Optional[Any] = None) -> dict:
```
aminalaee, line 945: I think instead you can do:

```py
async def on_model_change(self, data: dict, model: Any, is_created: bool) -> None:
        """Perform some actions before a model is created or updated.
        By default does nothing.
        """

async def after_model_change(
        self, data: dict, model: Any, is_created: bool
    ) -> None:
        """Perform some actions after a model was created or updated and committed to the database.
        By default does nothing.
        """
```

sqladmin/application.py
```diff
@@ -475,7 +475,8 @@ async def edit(self, request: Request) -> Response:
                 status_code=400,
             )
 
-        await model_view.update_model(pk=request.path_params["pk"], data=form.data)
+        form_ = model_view.on_model_change(data=form.data, pk=request.path_params["pk"])
+        await model_view.update_model(pk=request.path_params["pk"], data=form_)
```
aminalaee, line 479: And instead of changes in `application.py` I think you can add them in `_queries.py` so it's running in the same transaction:

```py
    def _update_sync(self, pk: Any, data: Dict[str, Any]) -> None:
        pk = get_column_python_type(self.model_view.pk_column)(pk)
        stmt = select(self.model_view.model).where(self.model_view.pk_column == pk)

        with self.model_view.sessionmaker() as session:
            obj = session.execute(stmt).scalars().first()
            anyio.from_thread.run(self.model_view.on_model_change, data, obj, False)
            obj = self._set_attributes_sync(session, obj, data)
            session.commit()
            anyio.from_thread.run(self.model_view.after_model_change, data, obj, False)

    async def _update_async(self, pk: Any, data: Dict[str, Any]) -> None:
        pk = get_column_python_type(self.model_view.pk_column)(pk)
        stmt = select(self.model_view.model).where(self.model_view.pk_column == pk)

        for relation in self.model_view._relations:
            stmt = stmt.options(joinedload(relation.key))

        async with self.model_view.sessionmaker() as session:
            result = await session.execute(stmt)
            obj = result.scalars().first()
            await self.model_view.on_model_change(data, obj, False)
            obj = await self._set_attributes_async(session, obj, data)
            await session.commit()
            await self.model_view.after_model_change(data, obj, False)
```

aminalaee: You can do the same for insert method.

aminalaee (changes requested):
Thank you for the contribution.

Perfect timing 😄  I was about to create a PR.

aminalaee:
You can take #343 as an example.

sqladmin/application.py
```diff
@@ -435,7 +435,6 @@ async def create(self, request: Request) -> Response:
                 context,
                 status_code=400,
             )
-
```
aminalaee, line 438: Please revert this.

sqladmin/_queries.py
```diff
@@ -122,8 +124,10 @@ async def _update_async(self, pk: Any, data: Dict[str, Any]) -> None:
         async with self.model_view.sessionmaker() as session:
             result = await session.execute(stmt)
             obj = result.scalars().first()
+            await self.model_view.on_model_change(data, obj, False)
             obj = await self._set_attributes_async(session, obj, data)
             await session.commit()
+            await self.model_view.after_model_change(data, obj, False)
```
aminalaee, line 130: Please do the same thing for inserts except that they should pass True.

aminalaee (changes requested):

aminalaee:
You can run `black .` to make the pipeline pass.

aminalaee (approved):
Thank you for your contribution 🎉

PR #346, opened 2022-10-01
Add py.typed for the package to ship its typing information

franciscorode:
As described in https://www.python.org/dev/peps/pep-0561/

Currently, the next warnings are showed when running mypy

```python
class UserAdmin(ModelAdmin, model=UserModel):
```

> error: Unexpected keyword argument "model" for "__init_subclass__" of "object"
> error: Class cannot subclass "ModelAdmin" (has type "Any")


aminalaee (approved):
Thank you for the PR.

PR #349, opened 2022-10-02
Improve items list UI

ischaojie:
Changes:
1. Move Select items per page after pagination.
2. Ensure alignment.
3. Ensure actions along with the checkbox.

Preview:
image

And another piece of advice: It's better to move the actions button(view/edit/delete) to the end?

### DIFF
--- a/sqladmin/templates/list.html
+++ b/sqladmin/templates/list.html
@@ -35,21 +35,7 @@
       </div>
     </div>
     <div class="card-body border-bottom py-3">
-      <div class="d-flex">
-        <div class="dropdown text-muted col-4">
-          Show 
-          <a href="#" class="btn btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            {{ request.query_params.get("pageSize") or model_view.page_size }}
-          </a>
-          items
-          <div class="dropdown-menu">
-            {% for page_size_option in model_view.page_size_options %}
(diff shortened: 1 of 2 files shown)


sqladmin/templates/list.html
```diff
@@ -164,6 +152,18 @@ <h3 class="card-title">{{ model_view.name_plural }}</h3>
...
         </li>
       </ul>
+      <div class="dropdown text-muted">
+        <a href="#" class="btn btn-sm btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          {{ request.query_params.get("pageSize") or model_view.page_size }} / Page
+        </a>
+        <div class="dropdown-menu">
+          {% for page_size_option in model_view.page_size_options %}
+          <a class="dropdown-item" href="{{ request.url.include_query_params(pageSize=page_size_option) }}">
+            {{ page_size_option }} / Page
```
aminalaee, line 162: My only comment is about the wording here. Can we do something like `Show <x> items`?
I think that would be more readable.

aminalaee (changes requested):
Thanks a lot for the contribution, The UI can be improved a lot.

sqladmin/templates/list.html
ischaojie: Yes, I'll add it.

aminalaee (approved):
Thanks for your contribution

PR #381, opened 2022-11-16
Fix export to use `list_query` option

villqrd:
This PR addresses the issue described in https://github.com/aminalaee/sqladmin/issues/380

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,6 @@ dependencies = [
   "mypy==0.982",
   "psycopg2==2.9.5",
   "pytest==7.2.0",
-  "pytest-asyncio==0.20.2",
   "sqlalchemy_utils==0.38.3",
   "sqlmodel==0.0.8",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ dependencies = [
   "mypy==0.982",
   "psycopg2==2.9.5",
   "pytest==7.2.0",
+  "pytest-asyncio==0.20.2",
   "sqlalchemy_utils==0.38.3",
   "sqlmodel==0.0.8",
 ]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -73,6 +73,7 @@ dependencies = [
   "mypy==0.982",
   "psycopg2==2.9.5",
   "pytest==7.2.0",
+  "pytest-asyncio==0.20.2",
```
aminalaee, line 76: I don't think this is needed, there's pytest anyio already used in other places. The same can be used here.

tests/test_models.py
```diff
@@ -57,6 +57,13 @@ def prepare_database() -> Generator[None, None, None]:
     Base.metadata.drop_all(engine)
 
 
+@pytest.fixture()
+def session(prepare_database) -> LocalSession:
```
aminalaee, line 61: Does it have to be a fixture? Can't it be instantiated within the test?

tests/test_models.py
```diff
@@ -483,3 +490,30 @@ class UserAdmin(ModelView, model=User):
...
+    batman = User(name="batman")
+    bruce = User(name="bruce wayne")
+    superman = User(name="superman")
+    session.add(batman)
+    session.add(bruce)
+    session.add(superman)
+    session.commit()
+    session.refresh(batman)
+    session.refresh(bruce)
+    session.refresh(superman)
```
aminalaee, line 506: It can simply work with one instance to prove the test. No need for the extra LOC.

aminalaee (changes requested):
Thanks a lot for the contribution

pyproject.toml
villqrd: indeed, i was wondering why your async tests were working and not mine, thanks for pointing out the secret! will change

tests/test_models.py
villqrd: see last changes

aminalaee (approved):
Thanks a lot for the contribution 🎉

PR #414, opened 2023-01-19
Support PostgreSQL arrays

aminalaee:
Fixes https://github.com/aminalaee/sqladmin/issues/412


aminalaee:
This works fine for the create page which stores data correctly, but in the edit form, the existing tags are not pre-populated:
https://github.com/aminalaee/sqladmin/blob/06f101741daef60c2a90d40cf98157ff25291996/sqladmin/application.py#L469

I expected the above code to populate select choices but it doesn't.

PR #440, opened 2023-03-06
Change AuthenticationBackend `authenticate` signature to support OAuth

aminalaee:
Closes https://github.com/aminalaee/sqladmin/issues/442

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -13,9 +13,10 @@ The class `AuthenticationBackend` has three methods you need to override:
 * `logout`: Will be called only for the logout, usually clearin the session.
 
 ```python
+from typing import Optional
+
 from sqladmin import Admin
 from sqladmin.authentication import AuthenticationBackend
-
 from starlette.requests import Request
 from starlette.responses import RedirectResponse
 
@@ -36,14 +37,13 @@ class AdminAuth(AuthenticationBackend):
         request.session.clear()
         return True
 
-    async def authenticate(self, request: Request) -> bool:
+    async def authenticate(self, request: Request) -> Optional[RedirectResponse]:
         token = request.session.get("token")
 
         if not token:
             return RedirectResponse(request.url_for("admin:login"), status_code=302)
 
-        # Check the token
-        return True
+        # Check the token in depth
 
 
 authentication_backend = AdminAuth(secret_key="...")
@@ -56,6 +56,8 @@ admin = Admin(app=..., authentication_backend=authentication_backend، ...)
 ??? example "Full Example"
 
     ```python
+    from typing import Optional
+
     from sqladmin import Admin, ModelView
     from sqladmin.authentication import AuthenticationBackend
     from sqlalchemy import Column, Integer, String, create_engine
@@ -91,7 +93,7 @@ admin = Admin(app=..., authentication_backend=authentication_backend، ...)
             request.session.clear()
             return True
 
-        async def authenticate(self, request: Request) -> RedirectResponse:
+        async def authenticate(self, request: Request) -> Optional[RedirectResponse]:
             if not "token" in request.session:
                 return RedirectResponse(request.url_for("admin:login"), status_code=302)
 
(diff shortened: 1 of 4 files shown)


simonsax:
Hi @aminalaee 

Thank you so much for your effort. I finally had time to test it (according to your gh-docs). But unfortunately it did't work for me.

- I'm not redirected to the auth_provider (in case no user is in the session) 
- I'm redirected to the admin backed regardless (if user is in session or not)

If I hit the login endpoint directly I get also this error message `authlib.integrations.base_client.errors.MismatchingStateError: mismatching_state: CSRF Warning! State not equal in request and response.`

as it is just one file I pasted the whole code below for reproduction (I only removed SSO credentials and installed fastapi, sqladmin  & authlib). 

In case I can help you, just let me know (but I'm not very experienced and learning day by day). And I'm definitely going to buy you a coffee for your help!

Do you spot the problem? Thanks and BR

```
from typing import Optional
from authlib.integrations.starlette_client import OAuth
from sqladmin.authentication import AuthenticationBackend
from starlette.applications import Starlette
from starlette.middleware.sessions import SessionMiddleware
from starlette.responses import RedirectResponse, Response
from fastapi import FastAPI

app = FastAPI()
app.add_middleware(SessionMiddleware, secret_key="test")

oauth = OAuth()
oauth.register(
    'google',
    client_id="...",
    client_secret="...",
    server_metadata_url="...",
    client_kwargs={
        'scope': 'openid email profile',
        'prompt': 'select_account',
    },
)
google = oauth.create_client('google')


Base = declarative_base()
engine = create_engine(
    "sqlite:///example.db",
)


class User(Base):
    __tablename__ = "users"

    name = Column(String)


class AdminAuth(AuthenticationBackend):
    async def login(self, request: Request) -> bool:
        return True

    async def logout(self, request: Request) -> bool:
        request.session.clear()
        return True

    async def authenticate(self, request: Request) -> Optional[RedirectResponse]:
        user = request.session.get("user")
        if not user:
            redirect_uri = request.url_for('login_google')
            return await google.authorize_redirect(request, redirect_uri)


admin = Admin(app=app, engine=engine, authentication_backend=AdminAuth("test"))


    def is_visible(self, request: Request) -> bool:
        return True

    def is_accessible(self, request: Request) -> bool:
        return True

admin.add_view(UserAdmin)

#also tried with @app.get('/auth/google')
@admin.app.route("/auth/google")
async def login_google(request: Request) -> Response:
    token = await google.authorize_access_token(request)
    user = token.get('userinfo')
    print(user)
    if user:
        request.session['user'] = user
    return RedirectResponse(request.url_for("admin:index"))
```

NikitaZakharov:
Hi @aminalaee 
Thanks for a great tool!

This doesn't look safe for me to authenticate user if return value is not `Response` type especially for devs that will upgrade from <0.10.0. 
Maybe if return value is not `Response` type then it's better to check that at least `bool(response) is True` ? 
It looks safer and at the same time is not a breaking change/vulnerability for devs who upgrade from <0.10.0.

aminalaee:
Hey @NikitaZakharov ,
Can you explain what is not safe? Maybe with a a sample code?
There's already a check doing `isinstance(resp, Response)` to make sure a response is returned.

NikitaZakharov:
@aminalaee 
So before it was pretty straightforward - to determine if user is authenticated we just check whether return value is true or false. Currently it's a requirement to return `Response` in order to properly handle authentication check. What bothers me is upgrading to 0.10.0 from previous versions might lead to always-authenticated issue. Also the idea that any non-Response return value triggers authentication check to be skipped makes me feel a bit unsafe. 

What if we just always expect `Response` type return value and otherwise just raise an error?

Current
```python
response = await auth_backend.authenticate(request)
if response and isinstance(response, Response):
    return response
# else we just skip authentication check and show admin page    
```

Suggested
```python
response = await auth_backend.authenticate(request)
if response and isinstance(response, Response):
    return response
raise Exception('Response instance expected')    
```

aminalaee:
hmm, yeah it is very implicit and the docs only mentions the Response object, but the code doesn't enforce it,
I think that's a good idea, feel free to create a PR with a simple test to cover it, otherwise I can take a look later.

PR #445, opened 2023-03-13
Fix bug to allow nullable zero values

ovginkel:
Bugfix: Previously incorrectly sets updates with 0 as value to None/NULL.

### DIFF
--- a/sqladmin/_queries.py
+++ b/sqladmin/_queries.py
@@ -44,8 +44,13 @@ class Query:
             relation = self.model_view._mapper.relationships.get(key)
 
             if not value:
-                # Set falsy values to None, if column is Nullable - but exclude False and zero values.
-                if not relation and column.nullable and value is not False and value is not 0:
+                # Set falsy values to None, if column is Nullable
+                if (
+                    not relation
+                    and column.nullable
+                    and value is not False
+                    and (value != 0 if isinstance(value, int) else False)
+                ):
                     value = None
 
                 setattr(obj, key, value)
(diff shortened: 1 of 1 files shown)


sqladmin/_queries.py
```diff
@@ -45,7 +45,12 @@ def _set_attributes_sync(self, session: Session, obj: Any, data: dict) -> Any:
 
             if not value:
                 # Set falsy values to None, if column is Nullable
-                if not relation and column.nullable and value is not False:
+                if (
+                    not relation
+                    and column.nullable
+                    and value is not False
```
aminalaee, line 51: Not really sure, but could use `bool(value)` to handle all falsy cases.
The downside is that there might be some types with unexpected behaviour.

aminalaee (commented):
Thanks for the PR. One small comment.

sqladmin/_queries.py
ovginkel: bool(None) still returns False instead of None making it indistinguishable from something that is false or zero and there are some variations in different handling by versions of python (e.g. empty arrays/mappings). Probably better to handle known cases explicitly to ensure compatibility across python versions. But I would be happy with any solution that allows me to set an integer field to 0 :)

aminalaee: Yeah probably you are right.
After thinking about this more and looking into the history, I think we can do something like:

```py
if not value:
                # Set falsy values to None, if column is Nullable
                if (
                    not relation
                    and column.nullable
                    and isinstance(value, bool)
                    and value is not False
                ):
                    value = None

                setattr(obj, key, value)
                continue
```

So we don't need to handle every type, the case that needs to be handled is to distinguish nullable booleans, otherwise we can set whatever value avaialble to the column.
And yes this works in your case of zero :)

sqladmin/_queries.py
```diff
@@ -78,7 +83,12 @@ async def _set_attributes_async(
...
+                    and (value != 0 if isinstance(value, int) else False)
+                ):
```
aminalaee, line 91:
```suggestion
                if (
                    not relation
                    and column.nullable
                    and isinstance(value, bool)
                    and value is not False
                ):
```

sqladmin/_queries.py
```diff
@@ -45,7 +45,12 @@ def _set_attributes_sync(self, session: Session, obj: Any, data: dict) -> Any:
...
```
aminalaee, line 53:
```suggestion
                if (
                    not relation
                    and column.nullable
                    and isinstance(value, bool)
                    and value is not False
                ):
```

aminalaee (approved):

PR #449, opened 2023-03-17
fix pk getters for related objects

timoniq:
i exposed a bug working with model of non-int pk \
you may use models i wrote in test suite to reproduce this bug

the problem was there:

https://github.com/aminalaee/sqladmin/blob/4eccf67f43fbafac19aa7373a9994942dfcc627f/sqladmin/models.py#L738-L760

these methods use `pk_column` of the current model instead of the related one \
so that the urls generated threw an error of undefined attribute (cause related model didn't obtain field with current model pk_column name) \
or here due to unproccessable `value` if the field names matched: https://github.com/aminalaee/sqladmin/blob/4eccf67f43fbafac19aa7373a9994942dfcc627f/sqladmin/models.py#L850

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -2,7 +2,7 @@ from typing import Generator
 
 import pytest
 from markupsafe import Markup
-from sqlalchemy import Boolean, Column, ForeignKey, Integer, String, Table, select
+from sqlalchemy import Boolean, Column, ForeignKey, Integer, String, select
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship, sessionmaker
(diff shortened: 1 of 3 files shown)


tests/test_models.py
```diff
@@ -431,6 +466,7 @@ class UserAdmin(ModelView, model=User):
     assert UserAdmin().get_export_columns() == [
         ("addresses", User.addresses.prop),
         ("profile", User.profile.prop),
+        ("permissions", User.permissions.prop),
```
aminalaee, line 469: Do any of these tests actually test this behaviour?

aminalaee (commented):
Thanks for the PR, just a minor comment.

tests/test_models.py
timoniq: perhaps some tests for url generations need to be added

timoniq: ok, i removed these tests

PR #454, opened 2023-03-24
feat: pluralize names with helper method

GriceTurrble:
Had an issue with models whose names already ends in 's'. While it makes sense in the app to either change the model names or implement name_plural directly, it also helps with the initial app usage if the model names are just a little cleaner earlier on. :)

As noted, the inflect package may be more useful there. I leave that to the maintainer's discretion.

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -87,6 +87,14 @@ def prettify_class_name(name: str) -> str:
     return re.sub(r"(?<=.)([A-Z])", r" \1", name)
 
 
+def pluralize(name: str) -> str:
+    # TODO introduce inflect package to do this more cleanly?
+    # https://pypi.org/project/inflect/
+    if name.lower().endswith("s"):
+        return name
+    return name + "s"
+
+
 def slugify_class_name(name: str) -> str:
     dashed = re.sub("(.)([A-Z][a-z]+)", r"\1-\2", name)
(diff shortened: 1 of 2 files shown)


sqladmin/helpers.py
```diff
@@ -87,6 +87,14 @@ def prettify_class_name(name: str) -> str:
     return re.sub(r"(?<=.)([A-Z])", r" \1", name)
 
 
+def pluralize(name: str) -> str:
+    # TODO introduce inflect package to do this more cleanly?
+    # https://pypi.org/project/inflect/
+    if name.lower().endswith("s"):
+        return name
+    return name + "s"
```
aminalaee, line 95: Shouldn’t the word end in es instead? I think that was the point!

```suggestion
    return name + "es"
```

aminalaee (commented):
Thanks for the PR, just a minor comment.

sqladmin/helpers.py
GriceTurrble: That depends! The inflect package would be better for (properly?) pluralizing words more dynamically.

We have some models like `Tasks` (improperly named, but that's a different topic) that in the original is pluralized `Taskss`. This would correct that issue.

aminalaee: Hmm..makes sense.
Can you share a sample code of your use-case, maybe in an issue?
I think it'd be nice to solve it without relying on another package first. The part that you dynamically create the views, I think can probably use inflect for that.

PR #463, opened 2023-04-13
Bugfix: Multiselect field saving only one value

nik-joseph:
Fixed issue where multiselect fields only select last value when creating/editing.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -2,10 +2,10 @@ import inspect
 from typing import (
     Any,
     Callable,
-    Dict,
     List,
     Optional,
     Sequence,
+    Tuple,
     Type,
     Union,
     no_type_check,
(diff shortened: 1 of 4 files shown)


sqladmin/application.py
```diff
@@ -594,22 +594,21 @@ async def _handle_form_data(self, request: Request, obj: Any = None) -> FormData
...
             if not isinstance(value, UploadFile):
-                form_data[key] = value
+                form_data.append((key, value))
                 continue
 
             should_clear = form.get(key + "_checkbox")
             empty_upload = len(await value.read(1)) != 1
             if should_clear:
-                form_data[key] = None
+                continue
```
aminalaee, line 606: I haven't checked this in details, shouldn't this be `form_data[key] = []` instead of `continue`?

nik-joseph: Changed type of `form_data` from `Dict` to `List[Tuple[str, Union[str, UploadFile]]]` 
so assuming you meant to change it to `form_data.append((key,[]))`
Ran a quick local test and none of the unitests have failed, so either there isn't a difference or tests aren't covering everything.
Can push the difference if you want.

aminalaee: yeah I think coverage is missing, I will add it in a separate MR.

aminalaee: Can we change it please?

nik-joseph: Just ran lint and I think its supposed to be `continue`. In starlette, the class is 
```
class FormData(ImmutableMultiDict[str, typing.Union[UploadFile, str]]):
    """
    An immutable multidict, containing both file uploads and text input.
    """

    def __init__(
        self,
        *args: typing.Union[
            "FormData",
            typing.Mapping[str, typing.Union[str, UploadFile]],
            typing.List[typing.Tuple[str, typing.Union[str, UploadFile]]],
        ],
        **kwargs: typing.Union[str, UploadFile],
    ) -> None:
        super().__init__(*args, **kwargs)
```

lint check fails with

1. `Argument 1 to "append" of "list" has incompatible type "Tuple[str, List[<nothing>]]"; expected "Tuple[str, Union[str, UploadFile]]` if I pass `[]`
2. `Argument 1 to "append" of "list" has incompatible type "Tuple[str, None]"; expected "Tuple[str, Union[str, UploadFile]]"` if I pass `None`

aminalaee (approved):
Thanks for the PR 🎉

nik-joseph:
> Thanks for the PR 🎉

Happy to help!

PR #464, opened 2023-04-13
Make empty uuids be none instead of empty strings

tobias-piotr:
## Description
When having a db model with nullable uuid field, the form is sending an empty string when the field is not filled. This makes postgres fail, because from the db point of view, uuid field can be filled with either a valid uuid or null.

### DIFF
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -196,6 +196,7 @@ async def test_model_form_postgresql() -> None:
     Form = await get_model_form(model=PostgresModel, engine=engine)
 
     assert len(Form()._fields) == 4
+    assert isinstance(Form()._fields["uuid"], NullableStringField)
     assert isinstance(Form()._fields["array"], Select2TagsField)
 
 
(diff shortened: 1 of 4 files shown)


tests/test_forms.py
```diff
@@ -278,3 +278,17 @@ class MyModel(Base):
...
+    class UUIDModel(Base):
+        __tablename__ = "uuid_model"
+
+        id = Column(UUID(as_uuid=True), primary_key=True)
+        team_id = Column(UUID(as_uuid=True))
+
+    Form = await get_model_form(model=UUIDModel, engine=engine)
+
+    assert len(Form()._fields) == 1
+    assert isinstance(Form()._fields["team_id"], NullableStringField)
```
aminalaee, line 294: Can we add the asserts to the existing test for postgres: test_model_form_postgresql ?

aminalaee:
Thanks for the PR.

tests/test_forms.py
piotr-tobiasz: done

PR #465, opened 2023-04-14
Fix nullable string fields

aminalaee:
Replaces https://github.com/aminalaee/sqladmin/pull/464


tobias-piotr (approved):
thank you for the fix, looks good

PR #473, opened 2023-04-18
reset UploadFile seek after reading

### LINKED ISSUE #472: UploadFile has had first byte read
### Checklist

### Describe the bug

Due to [this line](https://github.com/aminalaee/sqladmin/blob/main/sqladmin/application.py#L610), the file, when it gets to `model_view.insert_model()` or `model_view.update_model()` will be offset by one.

```python
async def insert_model(self, data: dict[str, Any]) -> Any:
        json_data_upload_file: UploadFile = data["json_data"]
        json_data_bytes = await json_data_upload_file.read()
        json_data = json.loads(json_data_bytes)
        # ...snip...
```

will always fail if the JSON starts with `{` or `[` or `"`, even if a valid JSON file was uploaded.

### Expected behavior

The file should not be left in an offset-seek state.

### Actual behavior

The file should arrives with a 1-offset.

### Debugging material

_No response_

### Environment

- sqladmin = "0.10.2"
- Python = "3.11.3"

### Additional context

_No response_

murrple-1:
Fixes #472

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -608,6 +608,7 @@ class Admin(BaseAdminView):
 
             should_clear = form.get(key + "_checkbox")
             empty_upload = len(await value.read(1)) != 1
+            await value.seek(0)
             if should_clear:
                 form_data.append((key, UploadFile(io.BytesIO(b""))))
             elif empty_upload and obj and getattr(obj, key):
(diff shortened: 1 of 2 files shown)


aminalaee:
Hey, thanks for taking a look.
I think I already handled this in another step:
https://github.com/aminalaee/sqlalchemy-fields/blob/c7147d2cd61efbfe96fc34d615f54f3dc6087de7/sqlalchemy_fields/storages/filesystem.py#L57

But in your use-case you want to load the upload file before saving, which also shouldn't fail.
Can you please update this test: `test_create_form_post`
to `assert user.file.open().read() == b"abc" ` to make sure nothing is changed.

murrple-1:
Sorry, I've done my best searching in the repo, and I can't find a test with that name. Which file?

aminalaee (approved):
Thanks for the PR 🎉

PR #501, opened 2023-05-23
Support sqlalchemy.sql.sqltypes.Uuid

dexter-dopping-ekco:
Hi there,

It appears that SQLAlchemy 2.0 introduces a new UUID type, `sqlalchemy.sql.sqltypes.Uuid`. This is similar to `sqltypes.UUID` (typed with all caps), except that it also works with databases that don't have a native uuid type, in which case it uses `CHAR(32)`.

https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_15/lib/sqlalchemy/sql/sqltypes.py#L3522-L3533

This seems to be the default type used when using declarative mappings with Python's `uuid.UUID` type.


aminalaee (approved):
Nice, thanks for the PR 👍

PR #503, opened 2023-05-23
Proposal: Each ModelView can now have actions with the same name/slug

### LINKED ISSUE #502: `action.name` should be unique per `ModelView`, not globally
### Checklist

### Describe the bug

Now that 0.11.0 has released, I began working on the actual feature for my job. It works well, but with 1 unintuitve (to me) drawback I didn't discover earlier.

If you have:
```python
...
class FooAdmin(ModelView):
...
    @action(name="duplicate")
    async def duplicate(request: Request) -> Response:
        return Response("Foo Duplicate")
        ...

class BarAdmin(ModelView):
...
    @action(name="duplicate")
    async def duplicate(request: Request) -> Response:
        return Response("Bar Duplicate")
```

then hitting `GET /admin/foo/actions/duplicate` will return `Foo Duplicate` and hitting `GET /admin/bar/actions/duplicate` will return `Foo Duplicate` also. This is because the `identity` is a `path_param`, not an actual namespace. I recognize the action `name` needs to be unique, but I think it should be unique per `ModelView`.

```python
from sqladmin import Admin, ModelView, action
from starlette.applications import Starlette
from starlette.responses import Response

from sqlalchemy import create_engine
from sqlalchemy import (
    Column,
    Integer,
    String,
)
from sqlalchemy.ext.declarative import declarative_base

test_database_uri_sync = "sqlite:///test.db?check_same_thread=False"
engine = create_engine(test_database_uri_sync)
Base = declarative_base()
app = Starlette()
admin = Admin(app, engine)


class Foo(Base):
    __tablename__ = "foos"

    name = Column(String(length=16))

class Bar(Base):
    __tablename__ = "bars"

    name = Column(String(length=16))


class FooAdmin(ModelView, model=Foo):
    column_list = [Foo.id, Foo.name]

    @action(name="duplicate")
    async def duplicate(self, request: Request) -> Response:
        return Response("Foo Duplicate")

class BarAdmin(ModelView, model=Bar):
    column_list = [Bar.id, Bar.name]

    @action(name="duplicate")
    async def duplicate(self, request: Request) -> Response:
        return Response("Bar Duplicate")

# Base.metadata.drop_all(engine)
# Base.metadata.create_all(engine)

admin.add_view(FooAdmin)
admin.add_view(BarAdmin)
```

### Expected behavior

Hitting `GET /admin/foo/actions/duplicate` should return `Foo Duplicate` and hitting `GET /admin/bar/actions/duplicate` should return `Bar Duplicate`.

### Actual behavior

Hitting `GET /admin/foo/actions/duplicate` will return `Foo Duplicate` and hitting `GET /admin/bar/actions/duplicate` will return `Foo Duplicate` also.

### Debugging material

_No response_

### Environment

- Ubuntu 22.04
- Python 3.11

### Additional context

_No response_

murrple-1:
Fixes #502

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -145,9 +145,9 @@ class BaseAdmin:
             view_instance = cast(ModelView, view_instance)
             self.admin.add_route(
                 route=func,
-                path="/{identity}/action/" + getattr(func, "_slug"),
+                path=f"/{view_instance.identity}/action/" + getattr(func, "_slug"),
                 methods=["GET"],
-                name=f"{view_instance.identity}-{getattr(func, '_slug')}",
+                name=f"action-{view_instance.identity}-{getattr(func, '_slug')}",
                 include_in_schema=getattr(func, "_include_in_schema"),
             )
 
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -782,8 +782,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
     def _url_for_action(self, request: Request, action_name: str) -> str:
         return str(
             request.url_for(
-                f"admin:{self.identity}-{action_name}",
-                identity=self.identity,
+                f"admin:action-{self.identity}-{action_name}",
             )
         )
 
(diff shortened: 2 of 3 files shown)


aminalaee (approved):
Great, thanks for the PR 👍

PR #507, opened 2023-05-30
Implement multi pk support

dexter-dopping-ekco:
Hello!

This pull request adds multi primary key support to SQLAdmin.

Multi pk support is implemented using a function to convert a model instance into a single "identifier" and another function to split an identifier string back into individual pk values.

If a model has a single primary key, then the identifier is simply that field's value, which should mean these changes are compatible with existing code. When a model has multiple primary keys, their values are joined together to a string separated by semi-colons. Escaping applied to the primary key values in case they contain a semi-colon.

https://github.com/dexter-dopping-ekco/sqladmin/blob/4928a47c2e4ddf19cd6ba53d5bc2f53dc4c0d830/sqladmin/helpers.py#L192-L202

The rest of the changes are pretty simple. Wherever the old code used a single primary key, use the helper functions to make the code work with multiple primary keys instead.

Some things that are specific to single primary keys were kept in case users of SQLAdmin rely on those things in their code. I also tried to make sure the queries don't significantly change, for instance `queries._get_to_many_stmt` constructs the SQL statement in the same way as before if the model has a single primary key.

The existing test suite succeeds and I added some new tests. Let me know what you think.

Thanks,
Dexter


sqladmin/models.py
```diff
@@ -786,8 +791,8 @@ def _url_for_action(self, request: Request, action_name: str) -> str:
             )
         )
 
-    def _get_pk(self, obj: Any) -> Any:
-        return getattr(obj, get_primary_key(obj).name)
+    def get_identifier(self, obj: Any) -> Any:
```
aminalaee, line 794: Isn't this redundant? Can't we use the `get_object_identifier` in all places instead?

tests/test_pk_helpers.py
```diff
@@ -0,0 +1,79 @@
+from typing import Any
+
+import pytest
```
aminalaee, line 3: Can we merge this into the `test_helpers.py` file?

aminalaee (commented):
Hey, thanks for the PR. This is a great improvement.
Some minor changes in first review, I will need to try and review it again.

sqladmin/models.py
dexter-dopping-ekco: Yes that's true. I did this mainly to mirror the original code's `_get_pk` function and because this function is being used inside of some of the HTML templates. ([sqladmin/templates/details.html](https://github.com/aminalaee/sqladmin/pull/507/files#diff-bc54d17435a60b52d052a84ddfc98cd65333343f2ff639d100d1a2028302d315L68), sqladmin/templates/list.html)

Should I get rid of this method and make `get_object_identifier` available in the templates like so?:

```patch
diff --git a/sqladmin/application.py b/sqladmin/application.py
index 8adf907..0eb71a4 100644
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -92,6 +92,7 @@ class BaseAdmin:
         templates.env.globals["zip"] = zip
         templates.env.globals["admin"] = self
         templates.env.globals["is_list"] = lambda x: isinstance(x, list)
+        templates.env.globals["get_object_identifier"] = get_object_identifier
 
         return templates
 ```

aminalaee: Hm..yeah that's ok, let's keep it as you have already.

dexter-dopping-ekco:
Hi Amin,

I pushed a bugfix for the `_set_many_to_one` function. This used the `local_columns` field which returns a `set`. A `set` is unordered but the code relied on the column order matching with that of the identifier. During my development/testing these happened to always match so I didn't catch the issue.

I'm now using `local_remote_pairs` along with a dict to lookup the values, instead of assuming there's a property that returns the FKs in the order of the corresponding PKs.

Apologies for the oversight.

dexter-dopping-ekco:
Thank you for merging!

PR #542, opened 2023-07-14
Set `sessionmaker` on `BaseAdmin`

rossmacarthur:
This is a small change which makes it easier to access the sessionmaker if you are adding your own custom routes.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -69,6 +69,25 @@ class BaseAdmin:
         self.title = title
         self.logo_url = logo_url
 
+        if isinstance(engine, Engine):
+            self.sessionmaker = sessionmaker(
+                bind=self.engine,
+                class_=Session,
+                autoflush=False,
+                autocommit=False,
+            )
+            self.async_engine = False
+        elif isinstance(engine, AsyncEngine):
+            self.sessionmaker = sessionmaker(
+                bind=self.engine,
+                class_=AsyncSession,
+                autoflush=False,
+                autocommit=False,
+            )
+            self.async_engine = True
+        else:
+            raise TypeError("invalid engine type")
+
         middlewares = middlewares or []
         self.authentication_backend = authentication_backend
         if authentication_backend:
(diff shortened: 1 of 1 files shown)


aminalaee:
Thanks for the PR.
Can you add a sample code to show how this change aligns with the use-case mentioned?

sqladmin/application.py
```diff
@@ -182,41 +184,70 @@ def _handle_expose_decorated_func(
...
+                    self.admin.add_route(
+                        "/login_oauth",
+                        self.login_oauth,
+                        methods=["GET", "POST"],
+                    )
 
-            admin.add_model_view(UserAdmin)
+                async def login_oauth(self, request: Request) -> Response:
+                    ...
+                    sessionmaker, _ = self.get_sessionmaker()
```
aminalaee, line 208: Why can't we access it like this?

```py
async def login_oauth(self, request: Request) -> Response:
    with self.sessionmaker() as session:
        ...
```

aminalaee: I think the reason I'm confused is that since we are using session maker instead of using a session,
our new route doesn't need a new session maker, a single session maker should be enough for all routes, right?
They can just get new sessions from the sessionmaker.

rossmacarthur: Are suggesting that we set a `sessionmaker` attribute on the `Admin` class? I am subclassing `Admin` not `View`. A `View` has a `sessionmaker` attribute but not the `Admin` class.

aminalaee: You're right I missed that part.

I think need to move this: https://github.com/aminalaee/sqladmin/blob/978cdd6587c7fc8345076393ae6774cddbdcd48c/sqladmin/application.py#L203

To the Admin here: https://github.com/aminalaee/sqladmin/blob/978cdd6587c7fc8345076393ae6774cddbdcd48c/sqladmin/application.py#L66

And in the view only do:

```py
view.sessionmaker = self.sessionmaker
view.async_engine = False
```

rossmacarthur: Good idea, I have updated it to do that

aminalaee (approved):
Thanks for the PR 🎉

PR #547, opened 2023-07-14
Pass request to model view methods

### LINKED ISSUE #540: Access request context for database queries
### Checklist

All of my models have a `created_at_by_user_uid` and `updated_at_by_user_uid` column. I want to be able to automatically set these fields based on the authenticated user. The authenticated user is available on the `Request` but the request is not available to the database queries since:
- the database connection is provided globally not at request context level
- there is no way to inject field values based on the request context

I think there are various ways to solve this, here are some options
- Pass `Request` to `on_model_change` and friends
- at the query level, `delete_model`, `insert_model`, `update_model` could be passed the `Request` model
- do it in the model view by passing `Request` to functions like `get_object_for_edit`, `get_object_for_details`
- at the db session level: e.g. allowing the construction of a custom database session based on the request

- I'm currently doing it at the form level by overriding `_handle_form_data` which is very gross and it has some limitations, I have to check if it is a "create" or "update" by looking at the path.
- I considered overriding all of the route handlers but they're quite large.

### Additional context

_No response_

rossmacarthur:
Fixes #540

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -480,7 +480,7 @@ class Admin(BaseAdminView):
             if not model:
                 raise HTTPException(status_code=404)
 
-            await model_view.delete_model(model)
+            await model_view.delete_model(request, model)
 
         return Response(content=str(request.url_for("admin:list", identity=identity)))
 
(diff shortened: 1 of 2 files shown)


aminalaee (approved):
Thanks for the PR 👍

PR #548, opened 2023-07-14
Update models.py : support multiple OR search terms

johanboekhoven:
This splits the search term into a list, then adds each search term to the search expressions list.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -1078,16 +1078,19 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
 
     def search_query(self, stmt: Select, term: str) -> Select:
         """Specify the search query given the SQLAlchemy statement
-        and term to search for.
+        and term or terms to search for.
         It can be used for doing more complex queries like JSON objects. For example:
 
         ```py
         return stmt.filter(MyModel.name == term)
         ```
         """
-        expressions = [
-            cast(prop, String).ilike(f"%{term}%") for prop in self._search_fields
-        ]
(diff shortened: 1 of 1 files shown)


sqladmin/models.py
```diff
@@ -1078,16 +1078,19 @@ class UserAdmin(ModelView, model=User):
...
         ```
         """
-        expressions = [
-            cast(prop, String).ilike(f"%{term}%") for prop in self._search_fields
-        ]
+        terms = term.split()
+        expressions = []
+        for term in terms:
+            expressions.extend(
+                [cast(prop, String).ilike(f"%{term}%") for prop in self._search_fields]
```
aminalaee, line 1092: I think this is a nice addition, but there's a big side-effect:
If you search for a a keyword like "a car" it will also include unwanted search results as it will do "a" OR "car".

Not sure what this will be for other use-cases.
So far I think you can just override `def search_query`  in your own models to accomplish this.

I think to make it more general purpose we can use something like https://querybuilder.js.org/ on the UI side and do the search with Ajax.

aminalaee (commented):
Thank you for the PR.
I think we can improve this to be a very useful feature.

sqladmin/models.py
johanboekhoven: I understand, I already had some AND OR text filter in there as well to create different search possibilities, but it get complicated quickly. So I kept it simple. But I see how that solves my case but probably not all cases. So I will do as you said and override the behaviour for my particular case. Thanks!

PR #557, opened 2023-07-20
Use url_for for the create form action

nathanmargaglio:
The current template ignores the root_path for the app.  By using the url_for function, the root_path is used when applied.

### DIFF
--- a/sqladmin/templates/create.html
+++ b/sqladmin/templates/create.html
@@ -6,7 +6,7 @@
       <h3 class="card-title">New {{ model_view.name }}</h3>
     </div>
     <div class="card-body border-bottom py-3">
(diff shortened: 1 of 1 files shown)


sqladmin/templates/create.html
```diff
@@ -6,7 +6,7 @@
       <h3 class="card-title">New {{ model_view.name }}</h3>
     </div>
     <div class="card-body border-bottom py-3">
-      <form action="{{ request.url }}" method="POST" enctype="multipart/form-data">
+      <form action="{{ url_for('admin:create', identity=model_view.identity) }}" method="POST" enctype="multipart/form-data">
```
aminalaee, line 9: Can you please do the same for the edit form?

aminalaee (commented):
Thanks for the PR

PR #560, opened 2023-07-21
fix: remove request.url usage in templates

GriceTurrble:
`request.url` is a handy tool to reach for to get the current URL, but it depends heavily on the configuration of the application. This can return the incorrect URL or incorrect scheme when used behind a proxy, for instance.

Instead of going the route of implementing middleware or other config settings to correct it, this change takes advantage of default browser behavior. An empty URL used for `href` or `action` is automatically filled with the full URL from the browser's context. This also allows relative URLs from the current location, including URLs that require query params (these can include other Jinja variables so long as we use the `urlencode` filter for safety).

### DIFF
--- a/sqladmin/templates/create.html
+++ b/sqladmin/templates/create.html
@@ -6,7 +6,8 @@
       <h3 class="card-title">New {{ model_view.name }}</h3>
     </div>
     <div class="card-body border-bottom py-3">
-      <form action="{{ request.url }}" method="POST" enctype="multipart/form-data">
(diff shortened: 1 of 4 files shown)


aminalaee:
Thank you for the PR.
I think we should either have this or https://github.com/aminalaee/sqladmin/pull/557
which fix it in different ways.

GriceTurrble:
#557 does seem like the right direction, actually. I'm seeing through the tests there are a lot other spots where this change wouldn't work out.

PR #573, opened 2023-08-08
Add searchable and sortable relation columns

cwieken:
I have implemented the feature to define foreign key columns (relation columns) to be used for searching and sorting in list view.

This should also cover discussion https://github.com/aminalaee/sqladmin/discussions/514.


anton-petrov:
@aminalaee 
I tested this pull request on my production service - everything works correctly, so you can safely merge this PR.

aminalaee (commented):
Thank you for the PR, but I think it'd be easier to discuss the approach first in the issue/discussion before doing it.
I think it's ideal if we could keep the config in the same one i.e. doing this:

```py
class UserAdmin(...):
    column_sortable_list = ["address.street"]
```

Instead of introducing a new config for this, but not sure how much effort it is so I will take look and let you know.

aminalaee:
This is already implemented with a different API.

PR #575, opened 2023-08-10
Fix docs: Cookbook, Using a request object

s1beria21:
Fix docs: Cookbook > Using a request object. Post is created by author_id.


aminalaee (approved):
Thanks for the PR 👍

PR #585, opened 2023-08-19
Implement feature to filter for multiple search words in list

cwieken:
I implemented a feature to allow searching for multiple search terms separated by space char in list view.
Quoted strings can be used to still search for a single search term.

For example searching for:
- John Doe retrieves all users that have either John or Doe in the name
- "John Doe" retrieves only users that have the exact string "John Doe" in the name.

This covers discussion https://github.com/aminalaee/sqladmin/discussions/543 by implementing OR search terms separated by spaced but also ensures that searching for terms which contain spaces is still possible using AND (by using quotes).

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -1,3 +1,4 @@
+import shlex
 import time
 from enum import Enum
 from typing import (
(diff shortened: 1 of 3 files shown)


anton-petrov:
@aminalaee 
I tested this pull request on my production service - everything works correctly, so you can safely merge this PR.

aminalaee:
I still think using a jQuery library for creating more complex filters is more useful, it is more work for sure, but we could replace this search API with another for Ajax search but allow combining AND/OR operations in any way the user wants.
For example: https://querybuilder.js.org/

PR #592, opened 2023-08-25
Search in list when typing

anton-petrov:
No need to click Search button, automatically search data afer 1 seconds delay.

### DIFF
--- a/sqladmin/statics/js/main.js
+++ b/sqladmin/statics/js/main.js
@@ -51,6 +51,17 @@ $(document).on('keypress', '#search-input', function (e) {
   }
 });
 
+// Init a timeout variable to be used below
+var timeout = null;
+// Search
+$(document).on('keyup', '#search-input', function (e) {
+  clearTimeout(timeout);
+  // Make a new timeout set to go off in 1000ms (1 second)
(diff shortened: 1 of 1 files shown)


sqladmin/statics/js/main.js
```diff
@@ -51,6 +51,17 @@ $(document).on('keypress', '#search-input', function (e) {
...
 
+// Init a timeout variable to be used below
+var timeout = null;
+// Search
+$(document).on('keyup', '#search-input', function (e) {
+  clearTimeout(timeout);
+  // Make a new timeout set to go off in 1000ms (1 second)
+  timeout = setTimeout(function () {
+    $('#search-button').click();
+  }, 1000);
```
aminalaee, line 62: I'd argue if 1 sec is a good number, but let's see if anyone has any issues with this.

aminalaee (approved):
Thanks for the PR

sqladmin/statics/js/main.js
anton-petrov: In fact, this value is obtained empirically based on my own experience. IMHO, 1 second is good enough.

PR #597, opened 2023-08-30
Fix 'itsdangerous' import error when not using Authentication Backend

GriceTurrble:
## Issue
When importing `sqladmin.Admin`:

```py
from sqladmin import Admin
```

...the application imports from `.authentication`: https://github.com/aminalaee/sqladmin/blob/724b9a966262876701c24cfa051aeb663d385d48/sqladmin/application.py#L34

Which in turn calls `starlette.middleware.sessions`: https://github.com/aminalaee/sqladmin/blob/724b9a966262876701c24cfa051aeb663d385d48/sqladmin/authentication.py#L6

And in starlette package, `middleware.session` tries to import `itsdangerous`: https://github.com/encode/starlette/blob/cb6d284877cd359a21a0f31ebf8103a7b285a848/starlette/middleware/sessions.py#L5

However, this package is *not* installed automatically: it's part of the "full" deps: https://github.com/encode/starlette/blob/cb6d284877cd359a21a0f31ebf8103a7b285a848/pyproject.toml#L37

Therefore, it's easily possible to install `sqladmin` and *not* this crucial sub-dependency, creating an error when trying to run the site:
image

## Solution
Either:
1. Mark `itsdangerous` as a named dependency here (which seems odd to name a sub-dependency just to alleviate this issue); or
2. Install `starlette[full]` as a main dependency, ensuring all dependencies are covered up front.

This PR is a simple change implementing the latter.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
   "Operating System :: OS Independent",
 ]
 dependencies = [
-  "starlette",
+  "starlette[full]",
   "typing_extensions>=4.0;python_version < '3.8'",
   "sqlalchemy >=1.4",
   "wtforms >=3, <4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
   "Operating System :: OS Independent",
 ]
 dependencies = [
-  "starlette[full]",
+  "starlette",
   "typing_extensions>=4.0;python_version < '3.8'",
   "sqlalchemy >=1.4",
   "wtforms >=3, <4",
--- a/sqladmin/authentication.py
+++ b/sqladmin/authentication.py
@@ -3,7 +3,6 @@ import inspect
 from typing import Any, Callable, Union
 
 from starlette.middleware import Middleware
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
   "Operating System :: OS Independent",
 ]
 dependencies = [
-  "starlette",
+  "starlette[full]",
```
aminalaee, line 29: The only problem I see with this is that it will make itsdangerous required for everyone and also include other dependencies from starlette[full] which is not ideal. I guess we can do two things:
- first make the SessionMiddleware import in authentication.py optional like this and raise an error if not installed in runtime:
```py
try:
    from starlette.middlewares import SessionMiddleware
Except ImportError:
    SessionMiddleware = None
```
- and as you suggested we can use starlette[full] but we add sqladmin[full] for this.

aminalaee (commented):
Thanks for the PR, I had missed this for a long time.

aminalaee (approved):
Thanks for the fix :+1:

PR #620, opened 2023-09-15
fix broken link in doc

YannickLeRoux:


aminalaee (commented):
Hi, can you share what is the issue with the current URL?
I can see that it is working already in http://localhost:8080/sqladmin/cookbook/optimize_relationship_loading/

YannickLeRoux:
Interesting... I am not sure how to run this app locally. But on Github I get a 404:

screenshot-580

Same when I try to follow the link in my IDE:

screenshot-581

But it works here, so maybe it is not an issue: https://aminalaee.dev/sqladmin/cookbook/optimize_relationship_loading/

aminalaee:
If you want to run this locally you can install `hatch` and do `hatch run docs:serve` to serve the docs locally.

But you are right, with your change it works both in mkdocs and also in Github file browser.

aminalaee (approved):
Thanks for the PR.

Issue #624, opened 2023-09-21
Customized sort_query


YarLikviD:
### Checklist

- [X] There are no similar issues or pull requests for this yet.

### Is your feature related to a problem? Please describe.

At the moment, the admin panel does not support sorting by objects, but this can be solved using the custom sort_query method. Similar to list_query, search_query.

### Describe the solution you would like.

You can move the sorting from the main list method to separate methods.
``` Python
    async def list(self, request: Request) -> Pagination:
        ...
        sort_fields = self._get_sort_fields(request)
        stmt = self.sort_query(stmt, sort_fields)
        ...      
```
A separate method for getting fields.
```Python 
    def _get_sort_fields(self, request: Request) -> List[Tuple[str, bool]]:
        sort_by = request.query_params.get("sortBy", None)
        sort = request.query_params.get("sort", "asc")

        if sort_by:
            sort_fields = [(sort_by, sort == "desc")]
        else:
            sort_fields = self._get_default_sort()

        return sort_fields
```
And an overridable method that can be used at your discretion. 
```Python
    def sort_query(self, stmt: Select, sort_fields: List[Tuple[str, bool]]) -> Select:
        for sort_field, is_desc in sort_fields:
            if is_desc:
                stmt = stmt.order_by(desc(sort_field))
            else:
                stmt = stmt.order_by(asc(sort_field))

        return stmt
```
The idea of custom methods solves many problems and gives great flexibility.
I haven't actually added anything, but it solves the problem with sorting by object.

aminalaee:
I think that's a good idea, I think even passing `request` instead of `sort_fields` makes sense maybe they can customize it more.

So:

```py
def sort_query(self, stmt: Select, request: Request) -> Select:
        for sort_field, is_desc in sort_fields:
            if is_desc:
                stmt = stmt.order_by(desc(sort_field))
            else:
                stmt = stmt.order_by(asc(sort_field))

        return stmt
```

PRs are welcome.

PR #625, opened 2023-09-21
Add customized sort query signature (#624)

YarLikviD:
Implementation Issue (#624)
I added request to the sort_query method. 
It would be possible to sort only by clicking on the column, but this would reduce the flexibility of customization.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -1060,6 +1060,8 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         A method that is called every time the fields are sorted
         and that can be customized.
         By default, sorting takes place by default fields.
+
+        The 'sortBy' and 'sort' query parameters are available in this request context.
         """
         sort_by = request.query_params.get("sortBy", None)
         sort = request.query_params.get("sort", "asc")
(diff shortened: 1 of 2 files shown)


sqladmin/models.py
```diff
@@ -954,6 +943,20 @@ def get_export_columns(self) -> List[str]:
             defaults=self._list_prop_names,
         )
 
+    def _get_sort_fields(self, request: Request) -> List[Tuple[str, bool]]:
```
aminalaee, line 946: I think this method can be inlined in the `sort_query`, or at least be public to be used so overriding the sort_query is made easy. But I think it's small enough to fit into the public `sort_query`.

aminalaee (commented):
Looks great, thanks for the PR, just a minor comment.

sqladmin/models.py
YarLikviD: Do you mean something like that? Only one method.
```py
    def sort_query(self, stmt: Select, request: Request) -> Select:
        """
        A method that is called every time the fields are sorted
        and that can be customized.
        By default, sorting takes place by default fields.
        """
        sort_by = request.query_params.get("sortBy", None)
        sort = request.query_params.get("sort", "asc")

        if sort_by:
            sort_fields = [(sort_by, sort == "desc")]
        else:
            sort_fields = self._get_default_sort()

        for sort_field, is_desc in sort_fields:
            if is_desc:
                stmt = stmt.order_by(desc(sort_field))
            else:
                stmt = stmt.order_by(asc(sort_field))

        return stmt
```

aminalaee (approved):
Thanks for the PR :+1:

PR #659, opened 2023-10-27
Fix docs: Cookbook form_ajax_refs definition

j123b567:
I really don't understand it deeply but the original code example for `form_ajax_refs` from the cookbook does not work.
It is not documented anywhere why, I just found the answer in `tests/test_ajax.py`.

 - There is no element `child` in the `Parent` class but there is a `children` relationship
 - `"order_by": ("id",)` just does not work (`sqlalchemy.exc.ArgumentError: ORDER BY expression expected, got ('id',).`) but `"order_by": "id"` works

It would be also fine to add `__str__` method to the example implementation of the class `Child` to be more clear about what is used in the list.


aminalaee (approved):

PR #664, opened 2023-11-03
Fix docs: icon not rendered in writing custom views

WiraDKP:
We should add `fa-solid` to properly render the fontawesome icon

Before:
before_docs

After:
after_docs


aminalaee (approved):
Thank you for the PR 👍

PR #665, opened 2023-11-03
Fix brand icon is not showing

WiraDKP:
Fix the missing brand webfont.
Also I've updated the FontAwesome CSS to the current stable free version (6.4.2)

Before:
telegram_before

After:
telegram_after


aminalaee (approved):
Thanks for the PR 👍

PR #670, opened 2023-11-13
Fix multiple PK model containing boolean values

### LINKED ISSUE #669: Object identifier values function not parsing correctly path parameters when a field is False
### Checklist

### Describe the bug

When trying to edit, delete or view a record which has a primary key containing a False value, a 404 error is thrown. This is due to the `def object_identifier_values(id_string: str, model: type) -> tuple:` function in line 28. https://github.com/aminalaee/sqladmin/blob/9e163b48e49bb52433475ebb9ca2f25b04c55ac6/sqladmin/helpers.py#L224-L229

Notice that `get_column_python_type`returns a Python type, which is `bool`, but when `part` is called with `bool` it has a value of `"False"`, which leads to the unexpected behavior, as `bool("")` is the only way to return `False` in Python using `bool`.

1.Create a model which contains a PK with a Bool value and other PK.
2. Create an admin view for that model with some records having false as part of the PK.
3. Try to access the admin view and the detail/edit view or deleting the record. You will get a 404 error.
 
image

### Expected behavior

You should get the correct record when looking for a primary key which contains a falsey value.

### Actual behavior

The record can not be found

### Debugging material

image
False is being cast to True

### Environment

- Python 3.11, MacOS, sqladmin==0.15.2

### Additional context

_No response_

ncarvajalc:
Closes #669

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -225,11 +225,9 @@ def object_identifier_values(id_string: str, model: Any) -> tuple:
     values = []
     pks = get_primary_keys(model)
     for pk, part in zip(pks, _object_identifier_parts(id_string, model)):
-        python_type = get_column_python_type(pk)
-        if python_type is bool and part == "False":
-            values.append(False)
-        else:
-            values.append(python_type(part))
+        type_ = get_column_python_type(pk)
+        value = False if type_ is bool and part == "False" else type_(part)
+        values.append(value)
     return tuple(values)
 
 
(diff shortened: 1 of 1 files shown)


sqladmin/helpers.py
```diff
@@ -225,7 +225,11 @@ def object_identifier_values(id_string: str, model: Any) -> tuple:
     values = []
     pks = get_primary_keys(model)
     for pk, part in zip(pks, _object_identifier_parts(id_string, model)):
-        values.append(get_column_python_type(pk)(part))
+        python_type = get_column_python_type(pk)
+        if python_type is bool and part == "False":
+            values.append(False)
+        else:
+            values.append(python_type(part))
```
aminalaee, line 232:
```suggestion
        type_ = get_column_python_type(pk)
        value = False if type_ is bool and part == "False" else type_(part)
        values.append(value)
```

aminalaee (changes requested):
LGTM, minor change.

aminalaee (approved):
Thanks for the PR 👍

PR #699, opened 2024-01-18
Advanced file view and file download

EnotShow:
Now if the database value is the file path and the file exists it will be shown as a downloadable file

Before:
image
After:
image

Changes  affect deteils.html and list.html


EnotShow:
Don't know why but hatch didn't work on my machine.
How to fix that?

image

PR #702, opened 2024-01-23
New feature: File download support

### LINKED ISSUE #701: Add file downloading
### Checklist

I want to have the possibility to download files from the admin panel

If string is filepath than you can download it

image

I try to find way to add custom templating engine but, there is no solution to add custom templating engine so I can't add this functionality by myself

### Additional context

I already have solutions to resolve my problem. And I made a pull request for it. Please check and give me feedback.

korolenkowork:
Now if the database value is the file path and the file exists it will be shown as a downloadable file

Before:
image
After:
image

Changes  affect templates: deteils.html, list.html


agn-7:
This PR solves the #701 issue.

EnotShow:
I decided that the user wants to have not only a download method but a reading method too. Now, besides the download method, there's an additional option to open files directly in a browser.

Additionally, I've addressed a security concern from the previous version. Instead of using file paths in the URL, which is insecure.

Now the result looks like:
image

Please let me know when you check my MR. I'm will grateful for the feedback

mmzeynalli:
@korolenkowork any updates?

mmzeynalli:
@korolenkowork can we merge this logic with @maxim-f1 PR #999 ? Let's make this not default, but as type_formatter for user to use

korolenkowork:
@mmzeynalli worked on it long time ago, I'll try to have a look on it on Friday and let you know

mmzeynalli:
Yeah, we were passive for some time. Please do so, and let us know. Thanks!

korolenkowork:
image

image

korolenkowork:
@mmzeynalli I made needed updates, work as a charm with @maxim-f1 PR #999

mmzeynalli:
Looks nice! I will review it this week.

docs/working_with_templates.md
```diff
@@ -164,24 +164,10 @@ Usage in templates:
...
-
-Usage in templates:
-
-```
-{% if value_is_filepath(value) %}
-    {{ value }} is file path
-{% else %}
-    {{ value }} is not file path
-{% endif %}
-```
```
mmzeynalli, line 184: Why this was removed?

sqladmin/application.py
```diff
@@ -484,6 +501,18 @@ async def http_exception(
...
             Route("/login", endpoint=self.login, name="login", methods=["GET", "POST"]),
             Route("/logout", endpoint=self.logout, name="logout", methods=["GET"]),
+            Route(
+                "/{identity}/{pk:path}/{column_name}/download/",
+                endpoint=self.download_file,
+                name="file_download",
+                methods=["GET"],
+            ),
+            Route(
+                "/{identity}/{pk:path}/{column_name}/read/",
```
mmzeynalli: maybe /view?

sqladmin/fields.py
```diff
@@ -395,12 +406,91 @@ def process_data(self, value: list | None) -> None:
...
 
+class CDNURLField(fields.StringField):
+    """
+    URL field for CDN or remote file links (https://...).
+
+    Use via ``form_overrides`` together with :func:`file_display_formatter`` in
+    ``column_formatters`` for list/detail pages.
+    """
+
+    widget = widgets.TextInput()
```
mmzeynalli, line 429: There is URLInput in wtforms

mmzeynalli (changes requested):

mmzeynalli:
Unlike list/details/edit/delete flows, _get_file does not check is_accessible, can_view_details, or check_can_view_details.
So admin with limited access can request file content from models/rows they should not be able to view, if they can guess identity/pk/column_name.

Also, if there are two columns with let's say CDN urls, one if formatted, other is hidden (i.e. not exposed to admin, or exposed and hidden), then once can again download the hidden files. We need to verify the column names as well in download and read file endpoints

mmzeynalli (approved):

PR #707, opened 2024-02-05
Move non-field-specific errors to top of edit and create forms

jonocodes:
# Issue

In the create and edit templates, error messages under all the fields. This is problematic if your model has many fields that extend beyond the bottom of the window. In that case there is no visible feedback of the error unless you happen to scroll to the bottom.

See discussion here: https://github.com/aminalaee/sqladmin/discussions/706

# Changes

Move error messages to the top of form.

# Screenshot of change

Screenshot 2024-02-05 at 5 36 00 AM


aminalaee (approved):
Thanks for the PR

PR #713, opened 2024-02-15
Fix sort by model attribute

aminalaee:
Fixes https://github.com/aminalaee/sqladmin/issues/688


sqladmin/models.py
```diff
@@ -727,6 +725,9 @@ def _build_url_for(self, name: str, request: Request, obj: Any) -> URL:
             pk=get_object_identifier(obj),
         )
 
+    def _get_prop_name(self, prop: MODEL_ATTR) -> str:
```
lorg, line 728: You're using that code in many other places, consider replacing them as well?

e.g. 
Screenshot 2024-02-17 at 18 30 29

lorg:
The code itself looks great, I think there's an opportunity to make a consistent change elsewhere, up to you. I don't know what types of code review you are looking for here, if my feedback is not appropriate let me know and I'll delete.

sqladmin/models.py
aminalaee: Thanks for pointing this out, as far as I checked none of them were dealing with this case and they were checking other conditions.

PR #722, opened 2024-02-29
Add field description to Create/Edit templates

ngaranko:
Add Field.desctiption as small text to Create/Edit pages.

WTForms provide `Field.description` parameter to be used as help text placeholder and this PR makes use of that.

Model View for testing might look like:

```
class TestView(ModelView, model=TestEntity):
    form_args = {
        "field_id": {"description": "Additional description."}
    }
```


aminalaee (approved):

PR #731, opened 2024-03-19
update many to many use select2 multiple

ElfenSterben:
change QuerySelectMultipleField use select2 multiple


mmzeynalli:
@ElfenSterben can you give info about the problem and the fix, providing screenshots please?

ElfenSterben:
Sorry, it's been a long time since I worked on the project using this library, and I've forgotten what my original purpose was at that time.

PR #742, opened 2024-04-10
Hide save and add another button from edit.html if can_create is False

MaximZemskov:
Fixed an issue where buttons for creating new items were still shown even when can_create was set to False, leading to a 403 error if clicked.


aminalaee (approved):
Thanks for the PR 👍

Issue #733, opened 2024-03-20
Add ModelView.form_query property to create form-filtering feature


lukeclimen:
### Checklist

- [X] There are no similar issues or pull requests for this yet.

### Is your feature related to a problem? Please describe.

In the list view for a model with relationships, I am able to override the `ModelView.list_query` property to customize the select query. This allows me to add filters to show results that are relevant. This functionality doesn't exist for the form view of a model. 

### Describe the solution you would like.

Adding an extendable `ModelView.form_query` would add another tool to solve the problems [laid out in the documentation](https://aminalaee.dev/sqladmin/cookbook/optimize_relationship_loading) for relationship loading. Assuming that it's possible, the default function can remain the same as the existing query to populate a form, with users able to override that property in their own models if they need to.

### Describe alternatives you considered

There is `ModelView.form_ajax_refs`, which allows a more restricted experience to `ModelView.list_query`. Where this breaks down for me is if I need to filter the select statement based on a property that isn't directly on the class described in the class. `ModelView.form_ajax_refs` allows me to specify fields on the given relationship. In the docs, the example is the `street` and `zip_code` fields on the `address` relationship. To extend the example, if the `street` property pointed to an object, and I wanted to filter this relationship based on a property of the `street` class, I would not be able to do so.

### Additional context

There was [a related issue](https://github.com/aminalaee/sqladmin/issues/578) that was raised in the past (now closed), where the user had created their own custom query within the `update_model` call. I think that the use case in this issue would have benefitted from a `ModelView.form_query` function to override as a more simplified solution.

There have also been two recent posts ([here](https://github.com/aminalaee/sqladmin/discussions/712) and [here](https://github.com/aminalaee/sqladmin/discussions/729)) in the Q&A discussion area about wanting to filter relationships in the form view.

lukeclimen:
@aminalaee if you are able to confirm that the proposed changes would be possible in the form view, I can create a PR to address this.

aminalaee:
I think this makes sense, can you please proceed with a basic implementation and tests to see how it looks and we work on it?

PR #745, opened 2024-04-12
New Feature: Add edit form query function

lukeclimen:
This PR addresses the proposed changes that I brought up in issue #733.

The changes here should allow users to customize the query that populate the edit form for their models, so that they can filter the relationships to show just the relevant information (see [this discussion post](https://github.com/aminalaee/sqladmin/discussions/712) for reference).

Because the default function behaviour requires the primary key of the object for the form, I changed the scope of this PR to address the edit form, instead of both the edit and create forms.

The example that I added to the docs in the `optimize_relationship_loading.md` file has been tested locally. I didn't add it to the project, as the `test_model.py` file doesn't have the Parent or Child classes. If you'd like me to add the test for the sake of completeness, I can do so.


lukeclimen:
@aminalaee can you review this PR when you have some free time?

aminalaee:
> @aminalaee can you review this PR when you have some free time?
Sure, I will do that today or tomorrow. Thanks for the PR 👍

sqladmin/models.py
```diff
@@ -807,11 +807,13 @@ async def get_object_for_details(self, value: Any) -> Any:
...
-    async def get_object_for_edit(self, value: Any) -> Any:
-        stmt = self._stmt_by_identifier(value)
+    async def get_object_for_edit(self, request: Request) -> Any:
+        stmt = self.edit_form_query(request)
 
-        for relation in self._form_relations:
-            stmt = stmt.options(joinedload(relation))
+        if type(self).edit_form_query == ModelView.edit_form_query:
+            # If edit_form_query fn hasn't been overridden, add all relationships
+            for relation in self._form_relations:
```
aminalaee, line 815: Can we move this part to the `edit_form_query`? That way we can avoid this condition which doesn't seem nice.

aminalaee (commented):
Thanks for the PR 👍  Looks great, just a minor comment.

sqladmin/models.py
aminalaee: I think we can, but I'll wait for your feedback before modifying the code.

aminalaee (approved):
Thanks for the PR 👍

aminalaee:
Thanks for the explanation. I think it should be up to the user to use this properly. The default behaviour workws, and they can override the method and make sure it ia working again. We cannot control what the User might do with the query.

The reason I changed your code was that it was relying on a private method and the public one was checking if it was overridden which again is an anti-pattern.

gpkc:
FYI this PR seems to break an example that is present on the website: https://github.com/aminalaee/sqladmin/issues/886

PR #748, opened 2024-04-14
All files and folders in the "templates" directory are moved to the "templates/sqladmin" folder.

### LINKED ISSUE #747: sqladmin + jinja templates conflict index.html
### Checklist

### Describe the bug

for html static files jinja use folder "templates" , but if i create index.html  sqladmin redirect to this page  (Strange behavior)

1)Create new project
2) pip install fastapi uvicorn sqladmin
3) create basic main.py with sqladmin
4) create folder templates/index.html   (this file used for main pages not admin panels)
5) open http://127.0.0.1:8000/admin    ->  redirect to index.html

### Expected behavior

don't use index.html for admin page.   maybe change to admin.html  

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

windows python venv fasapi uvicorn sqladmin pycharm 

### Additional context

_No response_

hasansezertasan:
Closes #747

I ran the test suite, looking good 🚀.
I "visually" tested it with the example code from [Working with Custom Views - SQLAlchemy Admin](https://aminalaee.dev/sqladmin/writing_custom_views/), works fine ☮️.


aminalaee (approved):
Thanks a lot for the contribution 👍

lesleslie:
First, thanks all. I appreciate you.

I understand why it may be desirable to move templates to a 'templates/sqladmin' folder, but am curious why you just don't initialize the package loader with the 'templates/sqladmin' like so:

```
def init_templating_engine(self) -> Jinja2Templates:
        templates = Jinja2Templates("templates")
        loaders = [
            FileSystemLoader(self.templates_dir),
            PackageLoader("sqladmin", "templates/sqladmin"),
        ]
```

then you wouldn't have to add the 'sqladmin' to the extends. For instance in index.html:

`{% extends "sqladmin/layout.html" %}`

My point being, that if I have a custom 'layout.html', or other html file, that I want shared between sqladmin and another application (a directory that doesn't start with 'sqladmin') I can't do that then without overriding every other template file in the package that references 'sqladmin/layout.html'. But when the index.html file extends is as such:

`{% extends "layout.html" %}`

the package loader would still be able to find  'templates/sqladmin/layout.html' but other loaders (in priority of choice before the package loader) would also be able to load a custom version of the file in whatever directories are in the loader's searchpath (ie 'templates/something_else/layout.html') without having to modify all of the other core template files accessed by the package loader.

I feel the way this is implemented actually hinders customization, and reusability, of templates because if you want to use a single custom template that is not in a 'sqladmin' directory (possible because it's shared with other applications), you have to then modify/override every other template in the package that references it - and then manually edit each one of those every time changes are made to the sqladmin codebase or default templates when you could just be using the default package templates instead. 

I think this should be re-opened for discussion. Please let me know if this makes sense or if there is anything else I can do or provide to clarify my point.

Thanks again.

hasansezertasan:
> but am curious why you just don't initialize the package loader with the 'templates/sqladmin'

If you check out the problem at #747, you'll understand why we went in that direction.

> I can't do that then without overriding every other template file in the package that references 'sqladmin/layout.html'.

You are right, there are some problems but they are going to be resolved when they are real pain in the ass (like Flask Admin).

Here is a solution recommendation for your scenario: https://github.com/aminalaee/sqladmin/discussions/762#discussioncomment-9427086

I also want to point out that this idea comes from Flask Admin, which has proved it's reusability end extendibility.

lesleslie:
Thanks for the reply!

I love this project because it stays as true to flask-admin as it can.

After 'correctly' updating my custom admin code, directory structure, and templates, I have come to conclusion that I no longer have issues, and have made my peace, with this change.

hasansezertasan:
> Thanks for the reply!

You're welcome 🙏.

> I love this project because it stays as true to flask-admin as it can.

I totally agree, I'm working on a Starlette adoption of Flask Admin, let me know if you are interested in, I'll share a link.

> After 'correctly' updating my custom admin code, directory structure, and templates, I have come to conclusion that I no longer have issues, and have made my peace, with this change.

😂, happy to hear that. @flask-admin people did a lot of good design decisions, years of experience right?. Even though I don't understand some parts of its source code, I left them as is. They might know better! 🚀🚀🚀.

justatrade:
Thanks for your project, first of all!

I've faced an issue using a 0.19 version after 0.16.1. I could not just create **_templates_** folder, move there original package files and customize them a bit. Using _templates_dir_ parameter doesn't help.

Is there any way to achieve such goals like localisation and customisation?

Thanks in advance.

aminalaee:
I think I didn't fully get your question. 
Why would you move original package files there? You should override them in your templates directory.

justatrade:
Because I'd like to use them with minimal changes)
BTW, I found a solution, not sure if it's right but it works.
1. admin = Admin(app, templates_dir="templates")
2. Created templates folder with subfolder sqladmin
3. Moved all the original html files from package to templates/sqladmin
4. Changed all the paths to html files inside html files by adding "sqladmin/" to the left of the file name.

This works fine for me, and now I successfully override base files.

aminalaee:
Yes that is correct. This was the intention of this PR to need `sqladmin` inside the templates. Feel free to update the docs if you feel like there's missing information.

aminalaee:
I hope this clarifies what you said: https://github.com/aminalaee/sqladmin/pull/817

Issue #751, opened 2024-04-29
Raised value error


BhuwanPandey:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Raised valueerror when I passed string value for queryparameter page and pagesize

### Steps to reproduce the bug

To reproduce this issue when I pass page=ab and pagesize=rm like 
/admin/user/list?page=ab&pageSize=rm  

### Expected behavior

When user pass any value other than integer 
what I expected to
1. Either raised error message like
{
  "error": {
    "page": "value is not a valid integer",
     "pagesize": "value is not a valid integer",
  }
}
OR
2. show the default value like page=1 and pagesize=10

### Environment

window 11 /python 3.11 /

### Additional context

 Based on the approach that I provided which you think is more appropriate. For me, first one give more sense. 
If you think , Is this the actual issue. 
If so, I will be happy to contribute on this issue.

aminalaee:
hey, what does page=rm mean? these are supposed to be numbers as the errors suggests.

BhuwanPandey:
> hey, what does page=rm mean? these are supposed to be numbers as the errors suggests.

Page and Pagesize supposed to be number but when user pass value of  page and pagesize to some string value other than integervalue
It says (Internal server Error)
https://sqladmin-demo.aminalaee.dev/admin/address/list?page=somevalue
https://sqladmin-demo.aminalaee.dev/admin/address/list?page=2&pageSize=somevalue

aminalaee:
Yeah I think defaulting to the first page is fine. Feel free to create a PR for that.

hasansezertasan:
I also think that these parameters should have default values.

@BhuwanPandey looking forward to see your PR!

aminalaee:
Closed by #752

PR #752, opened 2024-04-30
Validate page and pageSize query parameters

BhuwanPandey:
For page and Pagesize query paramter , Set the default value when it contain non integer value .

Fix issue: #751 

Page and PageSize set to default value when called as 
https:domainname/admin/address/list?page=somevalue
https:domainname/admin/address/list?page=2&pageSize=somevalue

and
when called as  
https:domainname/admin/address/list?page=num 
https:domainname/admin/address/list?pageSize=num
where num lies in  (0,-1,-2...)
here, page and pagesize take only positive integer

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -688,6 +688,8 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         self._custom_actions_in_detail: Dict[str, str] = {}
         self._custom_actions_confirmation: Dict[str, str] = {}
 
+        self.page_size_ = self.page_size
+
     def _run_query_sync(self, stmt: ClauseElement) -> Any:
         with self.session_maker(expire_on_commit=False) as session:
             result = session.execute(stmt)
@@ -746,6 +748,12 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
 
         return value
 
+    def get_query_param(self, request: Request, key: str, default: int) -> int:
+        try:
+            return int(request.query_params.get(key, default))
+        except ValueError:
+            return default
+
     async def count(self, request: Request, stmt: Optional[Select] = None) -> int:
         if stmt is None:
             stmt = self.count_query(request)
@@ -753,10 +761,15 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         return rows[0]
 
     async def list(self, request: Request) -> Pagination:
-        page = int(request.query_params.get("page", 1))
-        page_size = int(request.query_params.get("pageSize", 0))
-        page_size = min(page_size or self.page_size, max(self.page_size_options))
+        page = max(1, self.get_query_param(request, "page", 1))
+        page_size = self.get_query_param(request, "pageSize", 0)
+        page_size = min(
+            max(0, page_size) or self.page_size, max(self.page_size_options)
+        )
         search = request.query_params.get("search", None)
+
+        # display page_size on list.html
+        self.page_size_ = page_size
 
         stmt = self.list_query(request)
         for relation in self._list_relations:
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -25,6 +25,7 @@ from sqlalchemy.orm.exc import DetachedInstanceError
 from sqlalchemy.sql.elements import ClauseElement
 from sqlalchemy.sql.expression import Select, select
 from starlette.datastructures import URL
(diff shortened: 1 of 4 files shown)


BhuwanPandey:
@aminalaee This PR is ready for review

sqladmin/models.py
```diff
@@ -688,6 +688,8 @@ def __init__(self) -> None:
         self._custom_actions_in_detail: Dict[str, str] = {}
         self._custom_actions_confirmation: Dict[str, str] = {}
 
+        self.page_size_ = self.page_size
```
aminalaee, line 691: I think this is a bit unclear, maybe it's better to have a method `def get_page_size` instead?

sqladmin/models.py
```diff
@@ -746,18 +748,29 @@ def _default_formatter(self, value: Any) -> Any:
 
         return value
 
+    def get_query_param(self, request: Request, key: str, default: int) -> int:
```
aminalaee, line 751: And also `validate_page_number` as it is not for any query parameter but integer only.

BhuwanPandey: you mean that get_query_param should be replaced with validate_page_number

aminalaee: Yeah that also works.

sqladmin/models.py
aminalaee: I think you should get rid of `self._page_size` as we already have a `self.page_size`. In order to get rid of that I think a helper method will be useful.

BhuwanPandey: I think , defining any method,we need variable to track changeable value of pagesize.
but, What variable do you suggest for that?
In above scenario, by replaceing self._page_size with something else will resolve the case.

aminalaee: I don't think we have to store it in a variable? Just as it was previously working it's either the current value coming from the request parameter or the default self.page_size.

BhuwanPandey: In list.html line_number  (197)
 {{ request.query_params.get("pageSize") or model_view.page_size }} / Page
 
when we leave as it is,
on passing pagesize=somevalue
domainname/admin/address/list?pageSize=somevalue
 
 this somevalue is displayed like
 pagesize

aminalaee: yeah but you could call`{{ model_view.validate_page_size(request.query_params.get("pageSize")) or model_view.page_size }} / Page` right?

BhuwanPandey: Yes,
we can do that, as well we have to set default value and change it's structure like

 model_view.validate_page_size((request.query_params.get("pageSize")) or model_view.page_size),model_view.page_size)
 
but it looks quite messy

aminalaee: I think it's better than having `self.page_size` and `self.page_size_` and also the page size should be evaluated per request, we can't store the current page size on model_view. The `page_size` stored is the configuration which doesn't change per request, we shouldn't store per request page_size there.

BhuwanPandey: in list.html
          {{ model_view.validate_page_number((request.query_params.get("pageSize") or model_view.page_size),model_view.page_size)  }} / Page

in models.py
def validate_page_number(self, page_value: int|str, default: int) -> int:
        try:
            return int(page_value)
        except ValueError:
            return default
  
async def list(self, request: Request) -> Pagination:
        page = max(1,self.validate_page_number(request.query_params.get("page", 1), 1))
        page_size = self.validate_page_number(request.query_params.get("pageSize", 0), 0)
        
 After updating  , it looks like that
 If all the changes looks good to you
 I will commit the changes

BhuwanPandey:
@aminalaee  Again, It is ready for Review

aminalaee:
If I understand correctly you expect these URLs to fail. This PR is not deployed yet, page=-500 will throw 400 exception.

BhuwanPandey:
> If I understand correctly you expect these URLs to fail. This PR is not deployed yet, page=-500 will throw 400 exception.

  when i tested locally, this condition will not trigger either page = -500 or pagesize = -1
  but, when both negative value of page and pagesize passed at once this condition is trigger...
  
  application.py
      if (
            pagination.page * pagination.page_size
            > pagination.count + pagination.page_size
        ):
            raise HTTPException(
                status_code=400, detail="Invalid page or pageSize parameter"
            )

aminalaee (approved):

PR #769, opened 2024-05-14
Add more docs for overriding default tempates

jonocodes:
Elaborate on how to do the template override.

Comes from https://github.com/aminalaee/sqladmin/discussions/762


aminalaee (approved):

PR #770, opened 2024-05-15
Implement flash messages

jonocodes:
This creates the 'flash' function which can be called from any view. Then the next page load will display the message.

It also added messages to the 'create', 'edit', 'delete' functions.

Screenshot from 2024-05-15 13-02-22

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -17,6 +17,7 @@ from typing import (
     TypeVar,
 )
 
+from starlette.requests import Request
 from sqlalchemy import Column, inspect
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import RelationshipProperty, sessionmaker
@@ -24,6 +25,13 @@ from sqlalchemy.orm import RelationshipProperty, sessionmaker
 from sqladmin._types import MODEL_PROPERTY
 
 T = TypeVar("T")
+
+
+class AlertTypeEnum(enum.Enum):
+    success = "success"
+    info = "info"
+    warning = "warning"
+    danger = "danger"
 
 
 _filename_ascii_strip_re = re.compile(r"[^A-Za-z0-9_.-]")
@@ -309,3 +317,14 @@ def choice_type_coerce_factory(type_: Any) -> Callable[[Any], Any]:
 
 def is_async_session_maker(session_maker: sessionmaker) -> bool:
     return AsyncSession in session_maker.class_.__mro__
+
+
+def flash(request: Request, message: str, alert_type: AlertTypeEnum) -> None:
+    if "_messages" not in request.session:
+        request.session["_messages"] = []
+
+    request.session["_messages"].append({"message": message, "alert_type": alert_type.value})
+
+
+def get_flashed_messages(request: Request):
+    return request.session.pop("_messages") if "_messages" in request.session else []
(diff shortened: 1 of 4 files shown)


hasansezertasan:
Looks like a quite good implementation.

It's similar to `litestar`'s flash plugin and `starlette-flash` package implementation.

Using session to keep the flash messages in the flash bag on redirects is a great idea but the user could use a custom `SessionMiddleware` that depends on a key value databade like Redis which might cause conflicts. I believe instead of using sessions to store flash messages, we can directly use cookies with a customizable cookie name default to `admin_flash_messages` to store everything.

Or we can come up with a middleware that is specifically for flashes, something like `FlashMiddleware`.

sqladmin/helpers.py
```diff
@@ -26,6 +27,13 @@
 T = TypeVar("T")
 
 
+class AlertTypeEnum(enum.Enum):
+    success = "success"
+    info = "info"
+    warning = "warning"
+    danger = "danger"
```
hasansezertasan, line 34: Why limit the message type? I think this could be just a string.

hasansezertasan (changes requested):
Looks like a quite good implementation. Great job 👍.

It's similar to `litestar`'s flash plugin and `starlette-flash` package implementation.

Using sessions to keep the flash messages in the flash bag on redirects is a great idea but the user could use a custom `SessionMiddleware` that depends on a key-value database like Redis might cause conflicts. I believe instead of using sessions to store flash messages we can directly use cookies with a customizable cookie name default to `admin_flash_messages` to store everything.

Or we can come up with a middleware that is specifically for flashes, something like `FlashMiddleware` 🤓

Also I do think we should update `edit.html` and `create.html` to use flash bag instead of plain error key check.

sqladmin/helpers.py
jonocodes: These map to tabler alert types https://tabler.io/docs/components/alerts

hasansezertasan: That's a good point but I believe that this shouldn't be limited by an enumeration.

Even though flash messages have some basic categories, they could have custom categories.

WDYT @aminalaee?

jonocodes: Looks like django uses an enum https://docs.djangoproject.com/en/5.0/ref/contrib/messages/#adding-a-message

jonocodes:
> Using sessions to keep the flash messages in the flash bag on redirects is a great idea but the user could use a custom `SessionMiddleware` that depends on a key-value database like Redis might cause conflicts. I believe instead of using sessions to store flash messages we can directly use cookies with a customizable cookie name default to `admin_flash_messages` to store everything.

Yeah we can do this in cookies, it you guys think its better.

> Also I do think we should update `edit.html` and `create.html` to use flash bag instead of using plain `error` variable.

Yes.

sqladmin/helpers.py
hasansezertasan: Oh I see...

zbyte64:
I think this may have been an instance of perfection being the enemy of good.

Requiring this to be done with cookies changes the contract in many places that only expose the request object. Consider the auth interface, how difficult it would be for it to flash a message that their password will expire if it has to manipulate the response object to do so.

What's the unhappy path here if their session backend isn't compatible? Can we give them an alternative so that the vast majority of deployments isn't blocked on this feature?

PR #777, opened 2024-05-30
Fix edit_form_query documentation example

### LINKED ISSUE #733: Add ModelView.form_query property to create form-filtering feature

lukeclimen:
The documentation example give for the `edit_form_query` in the cookbook references an older implementation of the function. This updates the cookbook to provide a working example.

fixes #733

### DIFF
--- a/docs/cookbook/optimize_relationship_loading.md
+++ b/docs/cookbook/optimize_relationship_loading.md
@@ -72,8 +72,7 @@ class ParentAdmin(ModelView, model=Parent):
     def edit_form_query(self, request: Request) -> Select:
         parent_id = request.path_params["pk"]
         return (
-            super()
-            .edit_form_query(request)
+            self._stmt_by_identifier(parent_id)
             .join(Child)
             .options(contains_eager(Parent.children))
             .filter(Child.parent_id == parent_id)


lukeclimen:
Hi @aminalaee, I wanted to update the docs to match the updated version of the `edit_form_query`, so I've submitted a small PR that updates the cookbook example. I have local tests that I've used to validate that this works as expected, and can add a commit if you'd like them included as well!

aminalaee (approved):
Thanks for the PR

PR #785, opened 2024-06-29
[fix] async func

SNex21:


aminalaee:
I'm guessing this was a mistake, so I'm closing this.

PR #805, opened 2024-08-20
Allow multiple ajax sorts and changes to result size

mfriedy:
Allow the passing of multiple columns into the order_by field of `form_ajax_refs`, and also allow the passing of a limit parameter, rather than always using the default.

This allows configurations such as 

``` python

    form_ajax_refs = {
        "order": {
            "fields": ("country_id", "user_id", "order_date"),
            "order_by": ["user_id", "order_date"],
            "limit": 50,
        }
    }

```


aminalaee (changes requested):
Thanks for the PR. Can you please add some unit tests for your change?

aminalaee (approved):
Thanks for the PR

PR #814, opened 2024-09-09
fix: clamp page if it exceeds the maximum page

alex-lambdaloopers:
Hello,

This change was asked by a client complaining about the error when filtering and being on a high page. 

Imagine we have 200 rows to display. The `page_size` is 10, hence the maximum page we should access is 20.

Now imagine that we want to filter out some of the rows, using the `search` parameter. After the search, there are 100 rows to display only, so the new maximum page is 10.

This parameter is applied after typing, and the current page parameter does not change. If we are in a page greater than the new maximum, we will be prompted with the error             

```py
raise HTTPException(
                status_code=400, detail="Invalid page or pageSize parameter"
            )
```

This seems like unintended behaviour to me. In my view, trying to access a page higher than the max in this situation should not trigger an error, but instead display the last page of them all. 

This PR clamps the page to its maximum value if possible. 

There is a caveat: the url displays the page that you specified, even though the UI will show the max page. so for example, the URL could be:

http://my-app/admin/some-item/list?search=someval&page=15

But the UI would show that we are on page 10.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -810,6 +810,9 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         else:
             count = await self.count(request)
 
+        # Clamp page
+        page = min(page, max(1, count // page_size + 1))
+
         stmt = stmt.limit(page_size).offset((page - 1) * page_size)
         rows = await self._run_query(stmt)
 
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -810,9 +810,6 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         else:
             count = await self.count(request)
 
-        # Clamp page
-        page = min(page, max(1, count // page_size + 1))
-
         stmt = stmt.limit(page_size).offset((page - 1) * page_size)
         rows = await self._run_query(stmt)
 
--- a/sqladmin/pagination.py
+++ b/sqladmin/pagination.py
@@ -46,15 +46,16 @@ class Pagination:
 
         raise RuntimeError("Next page not found.")
 
+    def __post_init__(self):
+        # Clamp page
+        self.page = min(self.page, max(1, self.count // self.page_size + 1))
+
     def resize(self, page_size: int) -> Pagination:
         self.page = (self.page - 1) * self.page_size // page_size + 1
         self.page_size = page_size
         return self
 
     def add_pagination_urls(self, base_url: URL) -> None:
-        # Clamp page
-        self.page = min(self.page, max(1, self.count // self.page_size + 1))
-
         # Previous pages
         for p in range(self.page - min(self.max_page_controls, 3), self.page):
             if p > 0:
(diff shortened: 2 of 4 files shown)


aminalaee:
Thanks for the PR, I think it makes sense.
Just an idea that maybe we could redirect to URL so that the caveat is avoided and if User enters page=10000 manually they will be redirected to page 10.

alex-lambdaloopers:
agreed. I thought so too. One idea could be adding this:

```diff

    @login_required
    async def list(self, request: Request) -> Response:
        """List route to display paginated Model instances."""

        await self._list(request)

        model_view = self._find_model_view(request.path_params["identity"])
        pagination = await model_view.list(request)
        pagination.add_pagination_urls(request.url)


+        request_page = int(request.query_params.get("page"))
+        if request_page > pagination.page:
+            return RedirectResponse(request.url.replace(query=f"page={pagination.page}"))


```

in `application.py`. it feels clunky to me since we will need to execute `model_view.list` again.

Another option would be add some js on the list.html that would run the same comparison, replacing the url if needed. This also feels clunky to me.

Do you like any of those alternatives? Or if you have another idea I´m open to implementing it.

aminalaee:
I thought we already had that logic in here: https://github.com/aminalaee/sqladmin/blame/e5e7d5073f5bd8a2a48adf1a30b8df4ffafdc647/sqladmin/application.py#L450

Instead of just raising an error we want to redirect.

alex-lambdaloopers:
you are correct. ill edit the pr

alex-lambdaloopers:
> it shouldn't happen in a normal case

the redirect would happen in the case described in this pr: we are on page 20 and add a search, then the max page becomes 10. We will call the list method, then redirect, then call the list method again. If this is not a problem for you it isn't for me

aminalaee:
Yeah I think it's fine for now. I don't have a better solution and we can re-visit this later.

alex-lambdaloopers:
ok updated. let me know what you think of this.

sqladmin/models.py
```diff
@@ -810,6 +810,9 @@ async def list(self, request: Request) -> Pagination:
         else:
             count = await self.count(request)
 
+        # Clamp page
+        page = min(page, max(1, count // page_size + 1))
```
aminalaee, line 814: I don't think we need this.

sqladmin/pagination.py
```diff
@@ -52,6 +52,9 @@ def resize(self, page_size: int) -> Pagination:
         return self
 
     def add_pagination_urls(self, base_url: URL) -> None:
+        # Clamp page
+        self.page = min(self.page, max(1, self.count // self.page_size + 1))
+
```
aminalaee, line 57: And I suggest to define this instead in the top:

```py
def __post_init__(self):
        # Clamp page
        self.page = min(self.page, max(1, self.count // self.page_size + 1))
```

aminalaee (commented):
Minor comment, but I think this is a nice fix

sqladmin/models.py
alex-lambdaloopers: true, deleted

sqladmin/pagination.py
alex-lambdaloopers: looks better, edited

aminalaee (approved):
Thanks

PR #829, opened 2024-10-09
add json export format.

Vasiliy566:
Add json-format to export.


Vasiliy566:
@aminalaee review please

ZharkovMihail (approved):

sqladmin/models.py
```diff
@@ -1179,6 +1182,32 @@ async def generate(writer: Writer) -> AsyncGenerator[Any, None]:
...
+            for row in data:
+                if not first_row:
+                    yield ","
+                else:
+                    first_row = False
+                row_dict = {
+                    name: await self.get_prop_value(row, name)
+                    for name in self._export_prop_names
+                }
+                yield json.dumps(row_dict)
```
aminalaee, line 1201: Can we check length of data before loop and if there is more than one item set `separator=","` then return `yield json.dumps() + separator`
I want to avoid this branching in loop.

aminalaee (commented):
This is a great addition. Just a minor comment.

sqladmin/models.py
Vasiliy566: dont get it - smth linke that?

`async def generate() -> AsyncGenerator[str, None]:
        yield "["
        separator = "," if len(data) > 1 else ""
        first_row = True
        
        for row in data:
            if not first_row:
                yield separator
            else:
                first_row = False
            
            row_dict = {
                name: await self.get_prop_value(row, name)
                for name in self._export_prop_names
            }
            yield json.dumps(row_dict)
        
        yield "]"
`

Vasiliy566: oh i got it, now looks like you want

Vasiliy566: @aminalaee is it ok? i want to use it on my side asap) can we merge?

Vasiliy566:
add name convert to string.

aminalaee (approved):
Thanks for the contribution 👍

PR #830, opened 2024-10-11
fix: handle pagination issue when search results do not match current…

vv-m:
### Description

This PR fixes a pagination issue in SQLAdmin where performing a search from a page other than the first page would result in an error (“Invalid page or pageSize parameter”) if the search results had fewer pages than the current page.

### Changes
- Added logic to check the total number of pages after performing a search.
- If the current page exceeds the total available pages in the search results, the user is automatically redirected to the first page.
- This ensures smooth user experience when navigating between search results and pagination.

### Testing
- Verified that search works correctly when on different pages of the pagination.
- Verified that the user is correctly redirected to the first page if the current page number exceeds the available pages in the search results.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -803,18 +803,21 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
             stmt = stmt.options(selectinload(relation))
 
         stmt = self.sort_query(stmt, request)
-        
+
         if search:
             stmt = self.search_query(stmt=stmt, term=search)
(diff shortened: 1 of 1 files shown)


aminalaee (commented):
Can you please add a basic unit test to cover this?

alex-lambdaloopers:
Hi, I believe this is a duplicate of #814. The approach is different in that this change redirects you to the first page, whereas 814 redirects to the maximum allowed page.

aminalaee:
Yeah I think you are right. This can be closed then.

PR #841, opened 2024-10-22
Fix: Cannot update value of attribute with reserved name when it starts empty

waltton:
Hi all, first PR here 👋 .

I observed this bug on a case that I have a field with the name `data` and I was able to create a record and update it without any issues if the first time I create the record with a value that is not empty.
If the record is created with empty value  in the `data` field I cannot update it via admin.

Took me a while to understand the checks so I added a comment.

At the end, my solution was just removing the check `and getattr(obj, reserved_field_name, None)`.
I thought about adding an or condition and check if the attribute had a value on the `obj` or `form_data` but I didn't saw the point of it. Maybe I'm missing something.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -719,8 +719,10 @@ class Admin(BaseAdminView):
             reserved_field_name = field_name[:-1]
             if (
                 field_name in data
+                # ensures that the object do not contain the reserved plus '_'
+                # eg.: if the object has an attribute 'data_'
+                # it won't be mistaken for a wtform denormalize field
                 and not getattr(obj, field_name, None)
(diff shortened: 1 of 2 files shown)


waltton:
@aminalaee I reverted the fix so the case I'm hitting is clear on the tests I added.
Please take a look at https://github.com/aminalaee/sqladmin/pull/841/checks#step:8:105

This is the issue I'm trying to fix.

If the object being passed has a empty value, the field is not denormalized.

mmzeynalli:
This is actual bug, I still exists at version 0.24.0. @waltton do you want to fix it, or shall I fix it?

PR #872, opened 2025-01-14
add a way to sort table by index values

fmotaf:
Currently, when dealing with tables that contain a lot of rows, it is quite difficult to find a specific line.
In order to solve this, I've made this PR to allow us to sort these rows.

sqladmin_gif

### DIFF
--- a/sqladmin/statics/js/main.js
+++ b/sqladmin/statics/js/main.js
@@ -150,3 +150,36 @@ $(':input[data-role="select2-tags"]').each(function () {
     $(this).append(option).trigger('change');
   }
 });
+
+// Add sort icons dynamically to the table headers
+$('th').not('.w-1').each(function () {
+  const columnName = $(this).text().trim(); // Get the column name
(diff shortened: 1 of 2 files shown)


mmzeynalli:
@fmotaf could you generalize this and make it appliable for any field? Also, current code sorts only current page, you have to handle this on Python side

mmzeynalli:
@fmotaf have you tried to use `column_sortable_list`? If I understand your problem correctly, there is already a solution for it, so this PR is obsolete. Please let me know if I should reopen it.

fmotaf:
Hey @mmzeynalli sorry for the delay, yeah the column_sortable_list solves the thing, it was more of a frontend stuff to be more interactive and easy to sort. Anyways i've figured out it only involves using JS, i mean it's not necessary to reopen but i believe it's a good QoL feature, any question please LMK.

mmzeynalli:
We are always open to new features. However, Im having hard time to understand what exacly you want that is not existent with `columnt_sortable_list`. Could you eloberate more?

PR #874, opened 2025-01-22
[Feature] Check the available primary actions (edit, delete, view details) for each row on listing page.

maxim-f1:
# Description
SQLAdmin has a convenient system for accessing certain actions from CRUD. I decided to extend this functionality. Now you can define basic actions for a certain record in the list, with the help of events:
- **check_can_view_details**
- **check_can_edit**
- **check_can_delete** 

Example of using such functionality:

```python
class ExampleView(ModelView, model=Example):
    can_view_details = True
    can_edit = True
    can_delete = True

    column_list = (
        Example.id,
        Example.can_view_details,
        Example.can_edit,
        Example.can_delete
    )

    async def check_can_view_details(self, request: Request, model: Any) -> bool:
        """
        You can add a custom model attribute checker before view details.
        """
        if model.can_view_details is False:
            return False
        return self.can_view_details

    async def check_can_edit(self, request: Request, model: Any) -> bool:
        """
        You can add a custom model attribute checker before edit.
        """
        if model.can_edit is False:
            return False
        return self.can_edit

    async def check_can_delete(self, request: Request, model: Any) -> bool:
        """
        You can add a custom model attribute checker before delete.
        """
        if model.can_delete is False:
            return False
        return self.can_delete
```
The result is in the admin panel:

image

# Important issues
1. Seamless migration from older versions is provided, even customized **Admin** classes should not break.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -2,6 +2,6 @@
 blank_issues_enabled: false
 contact_links:
 - name: Questions
-  url: https://github.com/aminalaee/sqladmin/discussions
+  url: https://github.com/smithyhq/sqladmin/discussions
   about: >
     The "Discussions" forum is where you want to start. 💖
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: "actions/checkout@v3"
       - uses: "actions/setup-python@v4"
         with:
-          python-version: 3.9
+          python-version: 3.12
       - name: "Install dependencies"
         run: pip install hatch
       - name: "Build package & docs"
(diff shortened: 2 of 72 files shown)


mmzeynalli (approved):
LGTM

mmzeynalli:
@maxim-f1 I just realized that, this part is redundant:

```python
if hasattr(model_view, "check_can_view_details"):  # model_view always have `check_can_view_details`, as it is inherited
            pk = request.path_params.get("pk")
            if pk is None or not isinstance(pk, str):
                raise ValueError(
                    f'pk not found in request.path_params "{request.path_params}"'
                )
            model = await model_view.get_object_for_details(request)
            can_view_details_row = await model_view.check_can_view_details(
                request, model
            )
            if can_view_details_row is not True:
                raise HTTPException(status_code=403)
```

No need for PR, i will fix it somewhere in between. Just letting you know

PR #877, opened 2025-02-03
feat: add CSV import with streaming progress and row-level validation

### LINKED ISSUE #933: Add import button
### Checklist

Now there is an export button, but there is no import button, it would be convenient in local development so as not to recreate the same data, it would be possible to export, and then be able to import them.

_No response_

_No response_

### Additional context

_No response_

alserious:
Adds opt-in CSV import to the model list page. Users upload a UTF-8 CSV file; the server validates each row (WTForms + foreign keys), optionally persists valid rows in a single transaction with per-row savepoints, and streams progress/results as NDJSON.

Import is disabled by default (can_import = False) and gated by check_can_import(request).

### ModelView configuration

| Option | Default | Purpose |
|--------|---------|---------|
| `can_import` | `False` | Enable import UI + endpoint |
| `column_import_list` | list page columns / PK | Columns expected in CSV |
| `column_import_exclude_list` | — | Exclude columns from import |
| `max_import_file_size` | 5 MiB | Max upload size |
| `import_max_rows` | `0` (unlimited) | Max data rows per file |
| `max_reported_missed_rows` | `100` | Cap on detailed error payload |

image


mmzeynalli (changes requested):

sqladmin/application.py
```diff
@@ -629,6 +641,58 @@ async def export(self, request: Request) -> Response:
...
+            relation_mapper = relation.mapper.class_
+
+            relation_objs = await model_view.get_relation_objects(relation_mapper)
+            if not relation_objs:
+                continue
+            for relation_obj in relation_objs:
+                for row in data:
+                    n_row = []
+                    for value in row.getlist(relation_name):
+                        if value == str(relation_obj):
```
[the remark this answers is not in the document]
alserious: Yes, it seems like this is the only way, as long as we don't use a primary key list. If we remove the relationship import, it won't be a problem.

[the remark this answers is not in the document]
mmzeynalli: @alserious this is nice feature. Do you plan to fix it so we can merge it?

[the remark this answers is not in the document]
alserious: @mmzeynalli removed the import of relations, simple import works

sqladmin/_queries.py
```diff
@@ -230,6 +230,42 @@ async def _insert_async(self, data: dict[str, Any], request: Request) -> Any:
...
+    def _insert_sync_many(
+        self, data: list[dict[str, Any]], request: Request
+    ) -> list[Any]:
+        objs = []
+        with self.model_view.session_maker(expire_on_commit=False) as session:
+            for row in data:
+                obj = self.model_view.model()
+                anyio.from_thread.run(
+                    self.model_view.on_model_change, row, obj, True, request
+                )
```
mmzeynalli: I believe this part wont work, as `obj` is not correct object. Please check `_insert_sync` and how they create obj and pass it to these functions

sqladmin/_queries.py
```diff
@@ -230,6 +230,42 @@ async def _insert_async(self, data: dict[str, Any], request: Request) -> Any:
...
+                obj = self._set_attributes_sync(session, obj, row)
+                objs.append(obj)
+            session.add_all(objs)
+            session.commit()
+            for obj, row in zip(objs, data):
+                anyio.from_thread.run(
+                    self.model_view.after_model_change, row, obj, True, request
```
mmzeynalli: same

sqladmin/_queries.py
```diff
@@ -230,6 +230,42 @@ async def _insert_async(self, data: dict[str, Any], request: Request) -> Any:
...
+                )
+        return objs
+
+    async def _insert_async_many(
+        self, data: list[dict[str, Any]], request: Request
+    ) -> list[Any]:
+        objs = []
+        async with self.model_view.session_maker(expire_on_commit=False) as session:
+            for row in data:
+                obj = self.model_view.model()
```
mmzeynalli: same

mmzeynalli: Also, do as you did above, user might want to change something between inserts in `after_model_change`

sqladmin/helpers.py
```diff
@@ -182,6 +183,24 @@ def stream_to_csv(
     return callback(writer)  # type: ignore
 
 
+def parse_csv(
+    csv_content: bytes, columns: list[str], delimiter: str = ","
+) -> list[MultiDict]:
+    if csv_content[:3] == b"\xef\xbb\xbf":
+        csv_content = csv_content[3:]
```
mmzeynalli, line 190: ?

sqladmin/models.py
```diff
@@ -1090,6 +1135,13 @@ async def delete_model(self, request: Request, pk: Any) -> None:
     async def insert_model(self, request: Request, data: dict) -> Any:
         return await Query(self).insert(data, request)
 
+    async def insert_many_models(
```
mmzeynalli: insert_models_many

mmzeynalli (changes requested):

mmzeynalli (approved):

sqladmin/helpers.py
vahidzhe: Instead of manually stripping the BOM, we can just decode with the `utf-8-sig` codec, it removes the BOM if present and is a no-op otherwise:

```python
try:
    _csv_content = csv_content.decode("utf-8-sig").splitlines()
except UnicodeDecodeError as exc:
    raise ValueError("CSV file must be UTF-8 encoded.") from exc
```

This drops the extra `if`/slice and handles the same Excel-exported CSV case.

sqladmin/application.py
```diff
@@ -756,6 +795,502 @@ async def export(self, request: Request) -> Response:
...
+                }
+            )
+
+            for line_number, row in enumerate(data, start=2):
+                await anyio.sleep(0)
+                if await request.is_disconnected():
+                    return
+
+                processed += 1
+                form = Form(row)
```
vahidzhe: Boolean values don't survive a round-trip. For non-nullable Boolean columns the scaffolded form uses a plain `BooleanField`, whose `false_values` is `(False, "false", "")`  so a CSV cell containing `False` (exactly what the exporter writes), or `0` / `no` / `off`, all validate to `True`. `helpers._coerce_bool` already implements the correct semantics; either reuse it for boolean columns before feeding the form, or use a `BooleanField` subclass with a wider `false_values` for the import form. An export -> import -> compare test would catch this.

sqladmin/application.py
```diff
@@ -756,6 +795,502 @@ async def export(self, request: Request) -> Response:
...
+                if not form.validate():
+                    row_errors.update(form.errors)
+
+                row_data = {col: row.get(col) for col in model_view._import_prop_names}
+                form_data_dict = self._denormalize_wtform_data(
+                    form.data,
+                    model_view.model,
+                )
+
+                # Import forms may omit FK columns (when form_include_pk=False).
```
vahidzhe: Columns kept from `row_data` here bypass type coercion entirely, so a FK column like `user_id` reaches `setattr()` as the raw string `"5"`. `_validate_foreign_key_values` already coerces the value via `coerce_column_value()` for the existence check, but the coerced value is discarded. On PostgreSQL/asyncpg this fails with a type error for any non-string column that isn't present in the form (FKs, PKs, ints, dates). Suggest coercing here , `coerce_column_value(mapper.columns[name], value)` , and turning a coercion failure into a row-level error instead of an unhandled exception.

mmzeynalli (approved):

aminalaee:
Great work @mmzeynalli  🚀

PR #880, opened 2025-02-08
Added a red asterisk (*) in front of the field name, which is not nullable.

maxim-f1:
Added a red asterisk (*) in front of the field name, which is not nullable.


aminalaee (approved):
Always look forward to UI improvements 👍

PR #883, opened 2025-02-17
Add model convertors docs.

Vasiliy566:
I encountered a big problem when trying to figure out how to change the default for an empty JSONField on the create and edit pages - I would like to add a note to the documentation about the model_converter and their use.


Vasiliy566:
@aminalaee

aminalaee (approved):

PR #890, opened 2025-03-13
Do not join the same model twice or more in search_query

aleksei-udalov:
Hello, thank you for this great project.

When I execute the query described in tests/test_models.py:test_search_query I'm getting the following error:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) ambiguous column name: users.name
[SQL: SELECT count(*) AS count_1 
FROM (SELECT addresses.id AS id, addresses.name AS name, addresses.user_id AS user_id 
FROM addresses JOIN users ON users.id = addresses.user_id JOIN users ON users.id = addresses.user_id JOIN profiles ON users.id = profiles.user_id
WHERE lower(CAST(users.name AS VARCHAR)) LIKE lower(?) OR lower(CAST(profiles.role AS VARCHAR)) LIKE lower(?) ORDER BY addresses.id ASC) AS anon_1]
[parameters: ('%tt%', '%tt%')]

```
This also happens when using the Postgres:
`sqlalchemy.exc.ProgrammingError: (psycopg.errors.DuplicateAlias) table name "users" specified more than once`

And happens both in sqlalchemy 1.4 and 2.0.

So I tried to fix it.


aleksei-udalov:
Also there seems to be a problem with mkdocs-autorefs which causes CI to fall: https://github.com/mkdocstrings/autorefs/issues/68

mmzeynalli:
@aleksei-udalov  updates?

mmzeynalli:
Closing due to inactivity. Also, this PR is identical to #983.

PR #891, opened 2025-03-13
Bump versions for mkdocs and related

aleksei-udalov:
`hatch run docs:build` are falling like here: https://github.com/mkdocstrings/autorefs/issues/68
For example https://github.com/aminalaee/sqladmin/pull/890


aminalaee:
Sorry I saw this too late, I don't think it's needed anymore

PR #911, opened 2025-05-07
export csv/json in utf-8

alserious:
Exporting non-ASCII data get issues

image

~~`encoding = "utf_8_sig"` added because MS Excel requires bytes signature `"\xEF\xBB\xBF"`~~
set `encoding = "utf-8"`

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -153,10 +153,9 @@ class _PseudoBuffer:
     """
 
     encoding = "utf_8_sig"
-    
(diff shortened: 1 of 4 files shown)


aminalaee:
Hey, will you please take a look?

alserious:
> Hey, will you please take a look?

Yes, thank you. I changed `utf-8-sig` to `utf-8` think it's best practice. Tests passed local, but stuck in ci pipeline.

PR #922, opened 2025-05-27
customizable identity attribute for ModelView URL generation

### LINKED ISSUE #681: Incorrect URLs with multiple views on the same model
### Checklist

### Describe the bug

The usage case is to provide two different views with `list_query` and slightly different templates.
Having two views on the same model requires to provide a different `name` and `identity` to the second view to avoid a crash.

However, both links in the menu point to the same `ModelView`, so only one of the two views is usable.

Side note: if a `name` is omitted in the second view, we get a `TypeError` in `Menu.add` (called from `BaseAdmin._build_menu`), which could probably be caught in advance and yield a more meaningful exception, such as "Multiple views on the same model must have different names".

MVE:

```python
from sqlalchemy import Column, Integer, String, create_engine, select, Select, func
from sqlalchemy.orm import declarative_base, Session

from fastapi import FastAPI

Base = declarative_base()
engine = create_engine(
    "sqlite:///example.db",
)


class User(Base):
    __tablename__ = "users"

    name = Column(String)
    email = Column(String)


with Session(engine) as session:
    if session.query(User.id).count() == 0:
        session.add(User(name='Demo1', email='demo1@demo.org'))
        session.add(User(name='Demo2', email='demo2@bad_domain.org'))
        session.commit()


# ------------------------

app = FastAPI()
admin = Admin(app, engine)


class BadDomainUserAdmin(ModelView, model=User):
    name = 'Bad user'  # omitting this leads to a TypeError
    identity = 'bad_user'

    def list_query(self, request: Request) -> Select:
        return select(User).where(User.email == 'demo2@bad_domain.org')

    def count_query(self, request: Request) -> Select:
        return select(func.count()).select_from(User).where(User.email == 'demo2@bad_domain.org')


admin.add_view(UserAdmin)
admin.add_view(BadDomainUserAdmin)

```

### Expected behavior

There should be two distincts views with distincts URLs (or at least, I didn't find anything suggesting that this in not supported).

### Actual behavior

Both menu entries lead to `/admin/user/list`, instead of `/admin/bad_user/list`.

### Debugging material

This seems to be due to `identity` being overwritten by the slugified model name:
https://github.com/aminalaee/sqladmin/blob/751329c16960ce2257c66ad19509fc657af9e4ca/sqladmin/models.py#L89

Changing the line to
```python
 cls.identity = attrs.get("identity", slugify_class_name(model.__name__))
```
seems to fix it at least at first glance, however, `identity` is referenced both in the templates and in `models.py` as a parameter to build urls, so this is not sufficient. Indeed, the urls for user edititing and detailed display both point to `user` instead of `bad_user`.

### Environment

* Linux Mint 21.1 Vera
* python 3.10.12
* SQLAdmin 0.16.0

### Additional context

The intention was to provide a customized list view for certain items in the table (so, listing only, no detail/edit view) with a custom form widget to allow the user to perform an action on some of these items.

### LINKED ISSUE #961: Two admin views for the same table/model
### Checklist

So, I have `Vacancy` table:

```python
from enum import Enum
from sqlalchemy.orm import DeclarativeBase

class VacancyStatus(str, Enum):
    DRAFT = 'draft'
    ACTIVE = 'active'
    INACTIVE = 'inactive'
    ARCHIVED = 'archived'
    DELETED = 'deleted'
    PAUSED = 'paused'
    EXPIRED = 'expired'
    CLOSED = 'closed'

class Vacancy(DeclarativeBase):
    __tablename__ = 'company_vacancy'
    id: Mapped[int]
    status: Mapped[VacancyStatus] = mapped_column(String, default=VacancyStatus.ACTIVE)
```

I have many unverified draft vacancies, which are later processed one by one. As they are many, I want to have two views in admin: Vacancy and DraftVacancy. I planned to separate with overriding `list_query` function, however, as it is the same class, I get the same identity, so URL is the same for both views.

Enable to indicate identity as parameter. If not given, use `slugify_class_name(obj.__class__.__name__)`.

I had to create another class to bypass.

```python
class DraftVacancy(Vacancy):
    pass
```

And use a new view for this "new" class"
### Additional context

SQLAdmin: v0.21.0

x23n5902y:
Allow specifying a custom 'identity' attribute in ModelViewMeta for URL generation

- Enable using a customizable 'identity' attribute instead of default slugified model class name.
- This allows creating separate ModelView instances for the same model with distinct URLs, improving flexibility and organization.

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -737,8 +737,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
     def _url_for_delete(self, request: Request, obj: Any) -> str:
         pk = get_object_identifier(obj)
         query_params = urlencode({"pks": pk})
-        url = request.url_for(
-            "admin:delete", identity=self.identity)
+        url = request.url_for("admin:delete", identity=self.identity)
         return str(url) + "?" + query_params
 
     def _url_for_details_with_prop(self, request: Request, obj: Any, prop: str) -> URL:
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -92,7 +92,7 @@ class ModelViewMeta(type):
             )
 
         cls.pk_columns = get_primary_keys(model)
-        cls.identity = slugify_class_name(model.__name__)
+        cls.identity = slugify_class_name(attrs.get("identity", model.__name__))
         cls.model = model
 
         cls.name = attrs.get("name", prettify_class_name(cls.model.__name__))
@@ -738,8 +738,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         pk = get_object_identifier(obj)
         query_params = urlencode({"pks": pk})
         url = request.url_for(
-            "admin:delete", identity=slugify_class_name(obj.__class__.__name__)
-        )
+            "admin:delete", identity=self.identity)
         return str(url) + "?" + query_params
 
     def _url_for_details_with_prop(self, request: Request, obj: Any, prop: str) -> URL:
@@ -754,7 +753,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
     def _build_url_for(self, name: str, request: Request, obj: Any) -> URL:
         return request.url_for(
             name,
-            identity=slugify_class_name(obj.__class__.__name__),
+            identity=self.identity,
             pk=get_object_identifier(obj),
         )
 


mmzeynalli:
@x23n5902y could you fix this? I fails only on your patch, currently all tests pass

sqladmin/models.py
```diff
@@ -99,7 +99,7 @@ def __new__(mcs, name, bases, attrs: dict, **kwargs: Any):
             ) from exc
 
         cls.pk_columns = get_primary_keys(model)
-        cls.identity = slugify_class_name(model.__name__)
+        cls.identity = slugify_class_name(attrs.get("identity", model.__name__))
```
mmzeynalli, line 102: I think it would be better to do:

`attrs.get("identity", slugify_class_name(model.__name__))`

mmzeynalli (changes requested):
@x23n5902y could you also write tests for this? Like two modelviews of the same model, with different identity. Have different actions on each of them, and check they dont "see" each others actions.

mmzeynalli:
So, I found the problem. In this patch, 4 tests (same 2 tests in sync and async_view) fail:

```
FAILED tests/test_views/test_view_async.py::test_list_view_with_relations - assert '<a href="http://testserver/admin/address/details/1">(Address 1)</a>' in '<!DOCTYPE html>\n<html lang="en">\n\n<head>\n  <meta charset="UTF-8">\n  <meta name="viewport" content="width=device-... <script type="text/javascript" sr...
FAILED tests/test_views/test_view_async.py::test_detail_page - assert '<a href="http://testserver/admin/address/details/1">(Address 1)</a>' in '<!DOCTYPE html>\n<html lang="en">\n\n<head>\n  <meta charset="UTF-8">\n  <meta name="viewport" content="width=device-... <script type="text/javascript" sr...
```

This happens because, in this patch, `sqladmin/models.py:ModelView:_build_url_for` function has beem changed from

```python
return request.url_for(
            name,
            identity=slugify_class_name(obj.__class__.__name__),
        )
```

To

```python
return request.url_for(
            name,
            identity=self.identity,
        )
```

Which works for most of the time, EXCEPT when building URL to FK objects, i. e., object from other model (in this case Address), and correspondingly, other identity. We cannot keep it as it was, because if Address model has user-defined identity, this will fail again.

Upon looking for solutions and chatting with AI, I got this:

```python
# application.py:BaseAdmin:add_model_view
self._model_identity_map[view.model] = view.identity

# models.py
def _resolve_identity(self, obj: Any) -> str:
        if isinstance(obj, self.model):
            return self.identity
        return self._admin_ref._model_identity_map.get(
            obj.__class__, slugify_class_name(obj.__class__.__name__)
        )

def _build_url_for(self, name: str, request: Request, obj: Any) -> URL:
        return request.url_for(
            name,
            identity=self._resolve_identity(obj),
        )
```

It is kind of an ugly solution, but it works. I am conflicted to fix this: it is really nice to have feature (I have needed it myself), but solution is ugly, and kind of hard to maintain. I will let @aminalaee to decide.

PR #945, opened 2025-08-07
Implement OperationColumnFilter to filter String, Numeric, and UUID Types

chezou:
This PR implements column filter with text input. Similar with flask-admin or Active Admin, we will create a filter class to achieve filtering a specific column.

This is to address https://github.com/aminalaee/sqladmin/discussions/374 also follows up https://github.com/aminalaee/sqladmin/pull/906#issuecomment-3036554538

The difference between "Search" field is Search field aims to search all possible fields configured, while column filter allows to filter a specific column.

image

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -141,8 +141,10 @@ The options available are:
     if you don't want to specify all the columns manually. For example: `column_list = "__all__"`
 
 ### ColumnFilter
- 
-A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title`, `parameter_name`, `lookups` and `get_filtered_query` are required.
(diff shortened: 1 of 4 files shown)


wicol:
Cool, nice work! But, this approach with separate query args for operation and attribute ("?name=bat&name_op=contains") seems a bit cumbersome compared to the combined approach of django-filter: "name__contains=bat". Don't you think?
I'd say it even gets a little confusing when using operators like gt, isnull etc since the value supplied represents what you DON'T want to find:
* `?created_at=2025-01-01&created_at_op=gt`
* `?deleted_at=false&deleted_at_op=isnull`

Also it wouldn't hurt to use a familiar syntax for those with experience with django-filters (and maybe others?).

Just my 2c!

aminalaee (approved):

chezou:
@aminalaee Thanks for the review and merging it! Looking forward to move on to the next official version!

PR #949, opened 2025-08-26
docs - update the example ColumnFilter

proby-actvo:
The `lookups` method requires a fourth, `run_query`, argument.
The `get_filtered_query` method requires a fourth, `model`, argument.

Also updated the docs, a minor clarification, that the above two are methods not fields.

It also looks like my editor removed some extra whitespace on save, I hope that's ok.

I'm also willing, if there's interest, to update the column filter code to add a `BaseFilter(ABC)` base class for all filters. Something to help enforce API consistency. Roughly:

```python
class BaseFilter(ABC):
    """
    Base class for all filters.

    All column filters should inherit this class and implement the required methods.
    """
    @abstractmethod
    def lookups(self, request: Request, model: Any, run_query: Callable[[Select], Any]) -> List[Tuple[str, str]]:
        raise NotImplementedError

    @abstractmethod
    def get_filtered_query(self, query: Select, value: Any, model: Any) -> Select:
        raise NotImplementedError
```

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -141,8 +141,8 @@ The options available are:
     if you don't want to specify all the columns manually. For example: `column_list = "__all__"`
 
 ### ColumnFilter
- 
-A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title`, `parameter_name`, `lookups` and `get_filtered_query` are required.
+
+A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title` & `parameter_name` and the methods `lookups` & `get_filtered_query` are required.
 
 ```python
 class IsAdminFilter:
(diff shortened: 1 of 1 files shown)


docs/configurations.md
```diff
@@ -141,8 +141,8 @@ The options available are:
     if you don't want to specify all the columns manually. For example: `column_list = "__all__"`
 
 ### ColumnFilter
- 
-A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title`, `parameter_name`, `lookups` and `get_filtered_query` are required.
+
+A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title` & `parameter_name` and the methods `lookups` & `get_filtered_query` are required.
```
birddevelper, line 145: Hi @proby-actvo , Thanks for your contribution. I think this version can be more clear:


```suggestion
A ColumnFilter is a class that defines a filter for a column. A few standard filters are implemented in `sqladmin.filters` module. Here is an example of a generic ColumnFilter. Note that the fields `title` and `parameter_name`, and the methods `lookups` and `get_filtered_query` are all required in a filter class.
```

birddevelper (changes requested):

birddevelper (approved):

PR #952, opened 2025-08-31
extra blocks for templates allowing customization

birddevelper:
Currently, adding a component to any part of the templates requires copying all the inherited template markup into the child template.

This PR refactors the `content` block into smaller, modular blocks, similar to how Django admin templates are structured. This allows developers to seamlessly insert components into a page without duplicating the entire template.

For example, to add a new button next to the existing top buttons:

```html
{% extends "sqladmin/list.html" %}
{% block object_tools_items %}
{{ super() }}
<div class="ms-3 d-inline-block">
    <a href="some-link.html" class="btn btn-success">
        Some new button
    </a>
</div>
{% endblock %}
```
Result: 
image

This PR also splits the `head` block into two separate blocks, `extrahead` and `extrastyle` , similar to Django admin templates.

### DIFF
--- a/docs/working_with_templates.md
+++ b/docs/working_with_templates.md
@@ -28,23 +28,26 @@ SQLAdmin and in the `content` block it adds custom HTML tags:
 
 ## Overriding default templates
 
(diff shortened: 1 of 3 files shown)


mmzeynalli:
@birddevelper any updates with tests?

birddevelper:
Sorry, I couldn't add the tests yet. Can you please help adding tests?

PR #958, opened 2025-10-29
Fix date and time type handling when used as primary key

twoodwark:
Some scenarios call for a primary key that includes date or time types.

Fix the resulting exception "TypeError: 'str' object cannot be interpreted as an integer" when calling `object_identifier_values`.

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -6,7 +6,7 @@ import os
 import re
 import unicodedata
 from abc import ABC, abstractmethod
-from datetime import timedelta
+from datetime import date, datetime, time, timedelta
 from typing import (
     Any,
     AsyncGenerator,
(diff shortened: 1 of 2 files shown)


twoodwark:
@aminalaee when you have a moment please review this.

FWIW I've noticed a few other places where similar coercion is required using a column's python_type. It might be useful after landing this to think about refactoring.

sqladmin/helpers.py
```diff
@@ -226,7 +226,13 @@ def object_identifier_values(id_string: str, model: Any) -> tuple:
     pks = get_primary_keys(model)
     for pk, part in zip(pks, _object_identifier_parts(id_string, model)):
         type_ = get_column_python_type(pk)
-        value = False if type_ is bool and part == "False" else type_(part)
+        value: Any
+        if issubclass(type_, (date, datetime, time)):
+            value = type_.fromisoformat(part)
+        elif issubclass(type_, bool):
+            value = False if part == "False" else type_(part)
```
aminalaee: Maybe it should be this?
```suggestion
            value = False if part == "False" else True
```

PR #974, opened 2025-12-17
Migrate from hatchling to uv

mmzeynalli:
1. Migrate from hatchling to uv
2. Hatchling commands -> Makefile
3. Added more CI tests, covering both SA 1.4 and 2.0
4. Added bandit for security checks
5. Added pre-commit
6. Minor fixes to adapt to linters and checkers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,10 @@
 requires = ["uv_build>=0.9.17,<0.10.0"]
 build-backend = "uv_build"
 
+[tool.uv.build-backend]
+module-name = "sqladmin"
+module-root = ""
+
 [project]
 name = "sqladmin"
 version = "0.22.0"
@@ -76,15 +80,6 @@ docs = [
   "mkdocs==1.6.1",
   "mkdocstrings[python]==0.26.1",
 ]
-
-
-# -----------------------------
-# UV TOOL CONFIG
-# -----------------------------
-[tool.uv.build-backend]
-module-name = "sqladmin"
-module-root = ""
-
 
 [tool.mypy]
 disallow_untyped_defs = true


mmzeynalli:
@aminalaee could you retrigger tests? I do not think error was code problem

.github/workflows/test-suite.yml
```diff
@@ -9,12 +9,13 @@ on:
 
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
+    name: "Python ${{ matrix.python-version }} / SA ${{ matrix.sqlalchemy-version }}"
```
aminalaee:
```suggestion
    name: "Python ${{ matrix.python-version }} / SQLAlchemy ${{ matrix.sqlalchemy-version }}"
```

pyproject.toml
```diff
@@ -29,39 +28,24 @@ classifiers = [
 ]
 dependencies = [
   "starlette",
-  "sqlalchemy >=1.4",
   "wtforms >=3.1, <3.2",
   "jinja2",
   "python-multipart",
+  "sqlalchemy >= 1.4, <2",
```
aminalaee: We do want to publish for version 2

pyproject.toml
```diff
@@ -76,68 +60,72 @@ dependencies = [
   "itsdangerous",
   "phonenumbers",
   "pillow",
+  "psycopg2-binary",
```
aminalaee: Not sure why this was added, we are using psycopg[binary] the line after.

pyproject.toml
```diff
@@ -76,68 +60,72 @@ dependencies = [
...
+check_untyped_defs = true
+strict_optional = true
+no_implicit_optional = true
+disable_error_code = [
+  "var-annotated"
+]
+
+[tool.pylint.MASTER]
+disable = [
+  "missing-module-docstring",      # C0114
```
aminalaee: And here and in the code

aminalaee (commented):
LGTM, just a few tweaks and this will even be a samller PR

aminalaee:
@mmzeynalli You can change pytest coverage minimum, it's complaining now for 2% not being covered.

pyproject.toml
```diff
@@ -1,17 +1,16 @@
 [build-system]
-requires = ["hatchling"]
-build-backend = "hatchling.build"
+requires = ["uv_build>=0.9.17,<0.10.0"]
+build-backend = "uv_build"
```
aminalaee: I don't have any experience with UV build system, but it seems like to expect an `src` directory in the root:
https://docs.astral.sh/uv/concepts/build-backend/#modules

We might need to change that configuration as it describes with `module-root`

aminalaee:
```toml
[tool.uv.build-backend]
module-name = "sqladmin"
module-root = ""
```

PR #975, opened 2025-12-17
[Feature] Added switch style for checkbox and fixed related bug.

maxim-f1:
## Feature description
Updated the visual presentation of the checkbox. It now appears as a switch.

## Bug description
Also fixed a bug related to the checkbox. The bug was that the checkbox was considered a required field by default, but the required flag was not being passed into the flags list. Because of this, the field was not displayed as mandatory in the admin panel.

# Before
Screenshot after

# After
Screenshot before

## Backward Compatibility
* Compatible with all custom templates.
* Not applied when `form_args.{field_name}.widget` is overridden (e.g., `CustomWidget`).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,12 +4,16 @@ name: Publish
 on:
   push:
     tags:
-      - '*'
+      - "*"
 
 jobs:
   publish:
     name: "Publish release"
     runs-on: "ubuntu-latest"
+    permissions:
+      id-token: write
+      contents: write
+      attestations: write
 
     steps:
       - uses: "actions/checkout@v3"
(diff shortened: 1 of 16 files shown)


mmzeynalli (approved):

mmzeynalli:
Do we currently support `None` value for non-required boolean fields thou? If it is a checkbox/switch, how can user set `None` state from admin?

maxim-f1:
> Do we currently support `None` value for non-required boolean fields thou? If it is a checkbox/switch, how can user set `None` state from admin?

It also works with empty values, since it uses a selector widget.

https://github.com/aminalaee/sqladmin/blob/a0897e9e10e0bbff8fb5513f20f04855e004a7fb/sqladmin/forms.py#L371

image

mmzeynalli:
LGTM @aminalaee

mmzeynalli:
I have written to Amin multiple times through LinkedIn. He seems to be very busy, as I get no response. I have informed him that I will be creating a fork of this project, as I (and all of you) need new updates. I have merged this patch there. Please check https://github.com/mmzeynalli/sqladmin-ng.

PR #983, opened 2025-12-30
core: fix ambiguous column error when searching or sorting

nurikk:
Fixes SQLite "ambiguous column name" error raised when `column_searchable_list` or `column_sortable_list` references multiple fields from the same relationship (e.g. `["user.name", "user.email"]`).

### Cause
`_join_relationship_paths` joined via the target model class, which caused SQLAlchemy to emit duplicate JOINs across fields — and also failed with `AmbiguousForeignKeysError` when a model had multiple foreign keys to the same target (e.g. `billing_address` + `shipping_address` both pointing at `Address`).

### Fix
- Join via the relationship attribute (`stmt.join(Model.relation)`) so SQLAlchemy picks the right FK automatically.
- Track joined paths per call to avoid duplicates within a single query; SQLAlchemy dedupes joins on the same relationship attribute across calls, so the previous statement-introspection helper is no longer needed and has been deleted.
- Guard `search_query` against `or_()` being called with no arguments.

### Tests
- `test_search_multi_fields_no_duplicate_joins` — search alone with multiple fields from same relationship
- `test_sort_multi_fields_no_duplicate_joins` — sort alone with multiple fields from same relationship
- `test_sort_then_search_no_duplicate_joins` — cross-method dedup (sort then search)
- `test_search_two_fks_to_same_model` — model with two FKs to the same target
- `test_sort_and_search_together_no_ambigious_column_error` — HTTP-level regression guard

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -24,7 +24,7 @@ from typing import cast as typing_cast
 from urllib.parse import urlencode
 
 import anyio
-from sqlalchemy import Column, String, asc, cast, desc, func, inspect, or_
+from sqlalchemy import Column, String, asc, cast, desc, false, func, inspect, or_
 from sqlalchemy.exc import NoInspectionAvailable
 from sqlalchemy.orm import selectinload, sessionmaker
 from sqlalchemy.orm.exc import DetachedInstanceError
@@ -1200,9 +1200,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
             field_attr = getattr(model, parts[-1])
             expressions.append(cast(field_attr, String).ilike(f"%{term}%"))
 
-        if not expressions:
-            return stmt
-        return stmt.filter(or_(*expressions))
+        return stmt.filter(or_(false(), *expressions))
 
     def list_query(self, request: Request) -> Select:
         """


mmzeynalli:
@nurikk could you give me minimal code example in which the current code (without your patch) fails? I understand your description, but would like to test it myself as well.

nurikk:
> @nurikk could you give me minimal code example in which the current code (without your patch) fails? I understand your description, but would like to test it myself as well.

here you go:

```
from fastapi import FastAPI
from sqlalchemy import Column, ForeignKey, Integer, String, create_engine
from sqlalchemy.orm import declarative_base, relationship

# Database setup
Base = declarative_base()
engine = create_engine(
    "sqlite:///example.db",
)


class Author(Base):
    __tablename__ = "authors"

    first_name = Column(String)
    last_name = Column(String)
    

    books = relationship("Book", back_populates="author")

    def __str__(self):
        return self.name


class Book(Base):
    __tablename__ = "books"

    title = Column(String, nullable=False)
    year = Column(Integer)
    author_id = Column(Integer, ForeignKey("authors.id"), nullable=False)

    author = relationship("Author", back_populates="books")

    def __str__(self):
        return self.title


# App setup
app = FastAPI()
admin = Admin(app, engine)

class BookAdmin(ModelView, model=Book):
    column_list = [Book.id, Book.title, Book.year, Book.author]
    column_searchable_list = [Book.title, "author.first_name", "author.last_name"]
    column_sortable_list = [Book.id, Book.title, Book.year]

admin.add_view(BookAdmin)

```

so basically issue here:

`column_searchable_list = [Book.title, "author.first_name", "author.last_name"]`

results in 

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) ambiguous column name: authors.first_name
[SQL: SELECT count(*) AS count_1
FROM (SELECT books.id AS id, books.title AS title, books.year AS year, books.author_id AS author_id
FROM books JOIN authors ON authors.id = books.author_id JOIN authors ON authors.id = books.author_id
WHERE lower(CAST(books.title AS VARCHAR)) LIKE lower(?) OR lower(CAST(authors.first_name AS VARCHAR)) LIKE lower(?) OR lower(CAST(authors.last_name AS VARCHAR)) LIKE lower(?) ORDER BY books.id ASC) AS anon_1]
[parameters: ('%asd%', '%asd%', '%asd%')]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```

but if you allow search only by one field, it works fine

`column_searchable_list = [Book.title, "author.first_name"]`

sqladmin/models.py
```diff
@@ -1141,6 +1142,83 @@ class UserAdmin(ModelView, model=User):
...
+        Returns:
+            Set of joined entity classes
+        """
+        joined_entities: Set[Any] = set()
+
+        froms = stmt.get_final_froms()
+        if not froms:
+            return joined_entities
+
+        for from_clause in froms:
```
mmzeynalli: for from_clause in (froms or []):

tests/test_views/test_view_sync.py
```diff
@@ -855,3 +854,58 @@ def test_export_bad_type_is_404(client: TestClient) -> None:
...
+        user2 = User(name="Bob")
+        user3 = User(name="Charlie")
+        address1 = Address(user=user1)
+        address2 = Address(user=user2)
+        address3 = Address(user=user3)
+        session.add_all([user1, user2, user3, address1, address2, address3])
+        session.commit()
+
+    response = client.get("/admin/address/list?sortBy=user.name&sort=asc&search=o")
+    assert response.status_code == 200
```
mmzeynalli: test names are a little misleading. This names should be in test_models file, as you check for non-multiple joins. Here I would name `test_sort_and_search_together_no_ambigious_column_error`

mmzeynalli:
Also, there is such warning in tests:

```
tests/test_views/test_view_sync.py::test_search_multi_fields_no_duplicate_joins
tests/test_views/test_view_sync.py::test_sort_and_search_together_no_duplicate_joins
  /home/mzeynall/Projects/sqladmin/sqladmin/models.py:1209: SADeprecationWarning: Invoking or_() without arguments is deprecated, and will be disallowed in a future release.   For an empty or_() construct, use 'or_(false(), *args)' or 'or_(False, *args)'.
    return stmt.filter(or_(*expressions))
```

mmzeynalli:
@nurikk any updates? We would like to release new version soon.

sqladmin/models.py
nurikk: Removed the whole helper — joining via the relationship attribute lets SQLAlchemy dedupe across calls automatically.

tests/test_views/test_view_sync.py
nurikk: Moved the unit-level `no_duplicate_joins` tests to `test_models.py` and kept a single HTTP integration test here, renamed to `test_sort_and_search_together_no_ambigious_column_error`.

nurikk:
Addressed in latest commit:

- **Two FKs to same model** — fixed by joining via the relationship attribute (`stmt.join(Model.relation)`) instead of the target class, so SQLAlchemy picks the right FK. Dropped `_get_joined_entities` entirely since SQLAlchemy already dedupes joins on the same relationship attribute across calls. Regression covered by `test_search_two_fks_to_same_model`.
- **`or_()` deprecation warning** — `search_query` now returns early when no expressions are collected.
- **Test names/location** — unit-level dedup tests consolidated in `test_models.py`; `test_view_sync.py` keeps one HTTP-level test renamed to `test_sort_and_search_together_no_ambigious_column_error`.

sqladmin/models.py
```diff
@@ -1159,16 +1192,16 @@ def search_query(self, stmt: Select, term: str) -> Select:
...
-                stmt = stmt.join(model)
-
-            field = getattr(model, parts[-1])
-            expressions.append(cast(field, String).ilike(f"%{term}%"))
+            field_attr = getattr(model, parts[-1])
+            expressions.append(cast(field_attr, String).ilike(f"%{term}%"))
 
+        if not expressions:
+            return stmt
         return stmt.filter(or_(*expressions))
```
mmzeynalli: Let's do:

`return stmt.filter(or_(false(), *expressions))`

nurikk: Done in 01c220a.

Issue #990, opened 2026-01-12
"Save and continue editing" hidden wh


pperliti:
### Checklist

- [x] The bug is reproducible against the latest release or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Button "Save and continue editing" should be hidden in "create page" if `model_view.can_edit` is False.

### Environment

OS and Python: any
SQLAdmin version: 0.22.0

Issue #898, opened 2025-04-08
Unhandled exceptions during delete operations


savvan0h:
### Checklist

- [x] The bug is reproducible against the latest release or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

When attempting to delete a record that is referenced by another table via a NOT NULL foreign key, an IntegrityError is raised and results in an unhandled exception. This leads to a 500 Internal Server Error rather than a user-friendly message.

### Steps to reproduce the bug

Using the following minimal example:

```python
class RelatedItem(SQLModel, table=True):
    id: int | None = Field(default=None, primary_key=True)
    items: list["Item"] = Relationship(back_populates="related_item")


class Item(SQLModel, table=True):
    id: int | None = Field(default=None, primary_key=True)
    related_item_id: int = Field(foreign_key="relateditem.id")
    related_item: RelatedItem | None = Relationship(back_populates="items")
```

1. Create a `RelatedItem` and an associated `Item` .
2. Attempt to delete the `RelatedItem` from the admin panel.
3. A server error (500) occurs due to the NOT NULL constraint on `Item.related_item_id` .

### Expected behavior

An appropriate error message should be shown to the user, explaining that the record cannot be deleted due to existing related items.

### Actual behavior

A 500 Internal Server Error is returned, with no indication of the root cause or how to resolve it.

### Debugging material

```
INFO:     172.19.0.3:58660 - "DELETE /admin/related-item/delete?pks=1 HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1970, in _exec_single_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 924, in do_execute
    cursor.execute(statement, parameters)
  File "/usr/local/lib/python3.10/site-packages/psycopg/cursor.py", line 732, in execute
    raise ex.with_traceback(None)
psycopg.errors.NotNullViolation: null value in column "related_item_id" of relation "item" violates not-null constraint
DETAIL:  Failing row contains (1, null).

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 401, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
[...]

PR #991, opened 2026-01-12
Showing exceptions happened during delete in list page as per #898

mmzeynalli:
Also patched #990: hiding "Save and continue editing" button in create page, if `can_edit` is False


mmzeynalli:
@aminalaee

PR #992, opened 2026-01-12
Fix TypeError with UUID primary keys in issubclass check

### LINKED ISSUE #981: Regression: TypeError with UUID primary keys in 0.22.0: issubclass() arg 1 must be a class
### Checklist

### Describe the bug

sqladmin 0.22.0 throws `TypeError: issubclass() arg 1 must be a class` when accessing the detail view for models with UUID primary keys using `Annotated` types. This is a regression introduced in version 0.22.0 - version 0.21.0 works correctly.

1. Create a model with UUID primary key using Annotated:
```python
from uuid import UUID
from typing import Annotated
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column

class Base(DeclarativeBase):
    pass

class User(Base):
    __tablename__ = "users"
    id: Mapped[UUID] = mapped_column(primary_key=True)
```

2. Set up sqladmin with this model
3. Navigate to user list view (works fine)
4. Click to view user details: `/admin/user/details/{uuid}`
5. Error occurs

### Expected behavior

Detail view should load successfully, displaying the user record details as it did in sqladmin 0.21.0.

### Actual behavior

Application crashes with `TypeError: issubclass() arg 1 must be a class` when trying to access any detail view for models with UUID primary keys.

### Debugging material

<details>
<summary>Full Traceback</summary>
```
File ".../sqladmin/models.py", line 904, in get_object_for_details
    stmt = self.details_query(request)
File ".../sqladmin/models.py", line 1162, in details_query
    return self.form_edit_query(request)
File ".../sqladmin/models.py", line 1178, in form_edit_query
    stmt = self._stmt_by_identifier(request.path_params["pk"])
File ".../sqladmin/models.py", line 918, in _stmt_by_identifier
    values = object_identifier_values(identifier, self.model)
File ".../sqladmin/helpers.py", line 230, in object_identifier_values
    if issubclass(type_, (date, datetime, time)):
       ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: issubclass() arg 1 must be a class
```

</details>

**Root Cause:**
PR #958 added a date/time type check in `helpers.py:230` but doesn't handle `Annotated` types properly. When `type_` is a `typing._AnnotatedAlias` (not an actual class), `issubclass()` fails.

**Proposed Fix:**
```python
# sqladmin/helpers.py line ~230
if isinstance(type_, type) and issubclass(type_, (date, datetime, time)):
    # existing logic
```

### Environment

- sqladmin: 0.22.0
- Python: 3.13
- SQLAlchemy: 2.0.38
- pydantic: 2.10.6

### Additional context

This regression was introduced in PR #958 which aimed to fix date/time primary key handling. The fix works for date/time types but breaks UUID primary keys when using type annotations.

Workaround: Downgrade to sqladmin 0.21.0

S3wnkin:
- Add inspect.isclass() guard before issubclass() call

- Prevents TypeError when column.python_type returns type annotations

- Fixes #981

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -268,7 +268,8 @@ def get_column_python_type(column: Column) -> type:
             return str
 
     if get_origin(python_type) is not None:
-        python_type = get_args(python_type)[0]
+        args = get_args(python_type)
+        python_type = args[0] if args else str
 
     return python_type
 
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -4,8 +4,9 @@ from typing import Annotated, Any
 from unittest.mock import MagicMock, PropertyMock
 
 import pytest
-from sqlalchemy import Column, Date, DateTime, ForeignKey, Integer, String, Time, Uuid
+from sqlalchemy import Column, Date, DateTime, ForeignKey, Integer, String, Time
 from sqlalchemy.orm import declarative_base
+from sqlalchemy.sql.type_api import TypeDecorator
 
 from sqladmin.helpers import (
     get_column_python_type,
@@ -152,17 +153,31 @@ def test_catch_malformed_id():
     test_case("Johnson;7;A;Extra")
 
 
-class UUIDPKModel(Base):
-    __tablename__ = "uuid_pk_model"
-    id = Column(Uuid, primary_key=True)
+#########################################################################
+##################### get_column_python_type() tests ####################
+#########################################################################
+class IntBackedType(TypeDecorator):
+    """TypeDecorator where python_type raises but impl (Integer) returns int."""
+
+    impl = Integer
+    cache_ok = True
+
+    @property
+    def python_type(self):
+        raise NotImplementedError
+
+
+class IntBackedPKModel(Base):
+    __tablename__ = "int_backed_pk_model"
+    id = Column(IntBackedType, primary_key=True)
 
 
 def test_get_column_python_type_with_uuid_pk():
     """Regression #981: must not raise TypeError when python_type
     returns a type annotation instead of a plain class."""
-    pk = UUIDPKModel.__table__.c["id"]
+    pk = IntBackedPKModel.__table__.c["id"]
     result = get_column_python_type(pk)
-    assert callable(result)
+    assert result is int
 
 
 def test_get_column_python_type_annotated_type_no_typeerror():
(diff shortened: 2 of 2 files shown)


mmzeynalli:
@S3wnkin any updates on tests?

mmzeynalli:
@S3wnkin any updates?

mmzeynalli:
Also, I think we need to update `get_column_python_type` for this case; in case of NotImplemented error, we can get `impl` of the type and get origin of the type:
Something like this:

```python
except NotImplementedError:
        if hasattr(column.type, "impl"):
            try:
                python_type = column.type.impl.python_type
                if not inspect.isclass(python_type) and hasattr(python_type, '__origin__'):
                    return python_type.__origin__
                return python_type
            except NotImplementedError:
```

mmzeynalli:
@S3wnkin I will be taking over this PR tomorrow, please let me know if you are going to add tests

S3wnkin:
@mmzeynalli Can you check and let me know if you need any more tests. Thanks!

mmzeynalli:
@S3wnkin seems nice. Could you add FastAPI-Users GUID type to the tests as well? I think that is the main part that issue/PR was created. Just copy paste type from source to the test.

mmzeynalli:
For example, first test (`test_get_column_python_type_with_uuid_pk`) was never a problem, because it is not annotated type, it works with current main as well

sqladmin/helpers.py
```diff
@@ -253,14 +257,20 @@ def get_direction(prop: MODEL_PROPERTY) -> str:
...
+                python_type = column.type.impl.python_type
             except NotImplementedError:
-                ...
-        return str
+                return str
+        else:
+            return str
+
+    if get_origin(python_type) is not None:
+        python_type = get_args(python_type)[0]
```
mmzeynalli: `get_args` can return empty list for basic types (`get_args(int) == ()`), so 

```
args = get_args(python_type)
python_type = args[0] if args else str
```

tests/test_helpers.py
```diff
@@ -147,3 +150,45 @@ def test_case(ident):
...
+
+
+def test_get_column_python_type_not_implemented_no_impl():
+    """Falls back to str when python_type raises NotImplementedError
+    and there is no impl."""
+    mock_col = MagicMock(spec=["type"])
+    t = MagicMock(spec=["python_type"])
+    type(t).python_type = PropertyMock(side_effect=NotImplementedError)
+    mock_col.type = t
+    assert get_column_python_type(mock_col) is str
```
mmzeynalli: Add real annotated type just in FastAPI-User GUID field. From the source, copy just one type, and add it to the test cases. Also indicate that it is taken from that library.

PR #996, opened 2026-01-16
fix: add RootPathMiddleware for proper static file routing with root_…

### LINKED ISSUE #538: Adding root_path breaks admin page
### Checklist

### Describe the bug

I have multiple endpoints and admin panel. I want my endpoints to be in root path of `/api/v1/`, but admin to be just `/admin/`. So, I added `app_configs["root_path"] = f"/api/v{settings.APP_VERSION}"`, and admin panel breaks down.

```
app_configs: dict[str, Any] = {"title": "Itinerago API"}
app_configs["root_path"] = f"/api/v{settings.APP_VERSION}"

app = FastAPI(**app_configs)
app.include_router(auth_router, prefix="/auth", tags=["Auth"])
app.include_router(core_router, prefix="", tags=["Main Logic"])


admin = Admin(app, engine)

# admin.add_view-s here
```

Commenting out second line fixes admin, however, it does not give desired output.

### Expected behavior

APIs will have `/api/v1/` prefix, but, admin will not. Admin will load without a problem.

### Actual behavior

Admin breaks down completely:

image

and

image

Links are also broken:
In the first image, when you click one of the models it goes to: /api/v1/admin/model/list

image

### Debugging material

_No response_

### Environment

Windows, Python 3.10, Dockerized Fastapi
sqladmin version 0.12.0

### Additional context

_No response_

### LINKED ISSUE #741: Ajax query passes incorrect urn when admin mounted on sub fastapi app
### Checklist

### Describe the bug

Ajax request from browser doesn't consider urn which setted up by sub fastapi app.

1) Link admin to sub fastapi app:
```
from fastapi import FastApi
from sqladmin import Admin

app = FastApi()
subapp = FastApi()
app.mount('/suburn', subapp)

admin = Admin(subapp, base_url='/') 
```
2) Link any ModelView with relationship with admin which forces ajax request at select list on create page.
3) Click on select list on create page and try to type symbols.

### Expected behavior

Ajax should request to uri as: `.../suburn/ModelViewName/ajax/...`

### Actual behavior

Ajax request to uri: `.../ModelViewName/ajax/...` with skipping `suburn`.

P.S. Admin page routing is correct with sub app.

### Debugging material

_No response_

### Environment

OS:
MacOS 14.4.1

Python:
3.11.8

sqladmin==0.16.1
fastapi==0.110.1

### Additional context

I understand that suburn not passed to template ajax, but all another routing work correctly - it confused me.
(I think that we need additional param like `sub_url` separatly off `base_url` to specific behavior.)

### LINKED ISSUE #1012: Incorrect AJAX lookup URLs when parent app is not rooted at /
### Checklist

### Describe the bug

When your root (non-admin) app is rooted at /something, the admin app's AJAX lookup endpoint is at /something/admin/...
However, in the generated UI, the actual requests are sent to /admin/... and fail.
It's because the [code](https://github.com/aminalaee/sqladmin/blob/main/sqladmin/application.py#L241) that generates these URLs doesn't take into account any root path prefix:
```python
view.ajax_lookup_url = urljoin(
    self.base_url + "/", f"{view.identity}/ajax/lookup"
)
```

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

MacOS, Python 3.13, SQLAdmin 0.23.0

### Additional context

_No response_

JartanFTW:
…path

When FastAPI/Starlette is deployed with root_path configured (e.g., behind a reverse proxy), static file requests could fail with 404 because Starlette's internal routing couldn't properly resolve paths for nested mounts.

This middleware normalizes request paths by prepending root_path when it's set but not included in the request path, ensuring StaticFiles and other nested mounts work correctly.

Fixes #538

All pytest cases pass. I also tested that this fix works via a docker-compose deployment on [JartanFTW/sqladmin/tree/test/root-path-deployment](https://github.com/JartanFTW/sqladmin/tree/test/root-path-deployment)


JartanFTW:
For whatever reason this issue has disappeared even without the fix :joy: Not sure what's going on, or if this PR is necessary.

aminalaee (changes requested):
Can you please add a test to verify it's working as expected?

mmzeynalli:
@JartanFTW add a test which after redirecting to some page, check page url, and make sure that, root_path is in the URL

mmzeynalli:
@JartanFTW any updates? Could you also add support for starlette 1.0 which was recently released? You just need to change `pyproject.toml`:

```toml
dependencies = [
  "starlette==0.49.3,<0.50.0; python_version == '3.9'",
  "starlette>=0.50; python_version >= '3.10'",
  "wtforms >=3.1, <3.2",
  "jinja2",
  "python-multipart",
  "sqlalchemy >= 2.0",
]
```

JartanFTW:
It's on my todo list, just very busy at the moment and this is low priority for me. Apologies!
I'll try to get to it later today.

JartanFTW:
**Design decisions**                                                                                                         
                                                                                                                           
  1. **Middleware moved from inner to outer app** — The inner admin Starlette app receives root_path already appended with base_url by Starlette's Mount. This caused the middleware to prepend /api/v1/admin + /admin/... = double prefix. Moving it to the outer app means it sees the original root_path before any Mount processing.
  2. **Scoped to path_prefix** — Since the middleware is on the user's app, we only rewrite paths starting with base_url to avoid touching their routes. The check uses the combined root_path + base_url prefix to determine whether rewriting is needed, which also correctly handles the edge case where root_path equals base_url. User routes already work fine with root_path because they don't go through nested Mount routing.
  3. **Starlette dependency pinned by Python version** — starlette==0.49.3 for Python 3.9 (last supported version), starlette>=0.50 for 3.10+.
  
  **What the middleware does not cover**
  
  - **WebSocket scope** — Consistent with the rest of the library which is HTTP-only.
  - **Non-admin mounts** — By design. Only admin routes need the fix because they use nested Starlette Mount routing. Top-levelroutes handle root_path natively.
  
  **Open question: Starlette pin for Python 3.9**
  The current pin is starlette==0.49.3 for Python 3.9, as requested by @mmzeynalli. Should this be ~=0.49.3 instead? That would allow patch releases (e.g., a hypothetical 0.49.4 security fix) while still staying within the 0.49.x line that supports Python 3.9. ==0.49.3 locks users out of any future patch.

JartanFTW:
Ok now I'm actually done. Let me know how it looks

tests/test_root_path.py
```diff
@@ -0,0 +1,222 @@
...
+from starlette.responses import PlainTextResponse, Response
+from starlette.routing import Route
+from starlette.testclient import TestClient
+
+from sqladmin import Admin, ModelView
+from tests.common import sync_engine as engine
+
+session_maker = sessionmaker(bind=engine)
+
+Base = declarative_base()  # type: ignore
```
mmzeynalli: I know this is copied, but no need for type: ignore. I will remove others as well.

mmzeynalli (approved):

mmzeynalli:
> **Open question: Starlette pin for Python 3.9** The current pin is starlette==0.49.3 for Python 3.9, as requested by @mmzeynalli. Should this be ~=0.49.3 instead? That would allow patch releases (e.g., a hypothetical 0.49.4 security fix) while still staying within the 0.49.x line that supports Python 3.9. ==0.49.3 locks users out of any future patch.

yes, totally agree. `~` us bettter than `==`.

mmzeynalli (approved):

mmzeynalli:
@JartanFTW would @MaximDementyev's [solution](https://github.com/smithyhq/sqladmin/issues/538#issuecomment-4202189863) work, instead of this much change?

JartanFTW:
@mmzeynalli That solution works when root_path is set explicitly in the constructor(FastAPI(root_path="/api/v1")), but not when it's configured on the ASGI server (uvicorn --root-path /api/v1). In that case, root_path is set per-request in the ASGI scope, not on the app itself, so self.app.root_path is "" at startup when add_view() runs, and the urljoin produces the same broken URL.
The fix I included uses url_for() at template render time, which reads root_path from the live request scope, so it works regardless of how root_path is set.

MaximDementyev:
It seems to me that it would be correct to separate what is being done and for what purpose.
1. There is a problem that ajax does not work if the application is behind a proxy with a root_path
2. There is a need to connect Admin to /admin and endponts via /api/v1

Possible solution:
1. The current MR solves ajax, but Middleware is not needed for this. Remove it and leave only the template changes
2. This problem is solved by creating a proper application tree. It might be worth adding an example to the documentation.

@JartanFTW , what other use cases am I missing? Your example https://github.com/smithyhq/sqladmin/issues/538#issuecomment-4201832301 you can solve this by combining points 1 and 2. Can you give an example of where you can't do without middleware?

JartanFTW:
@MaximDementyev

There are two ways `root_path` gets set, and they behave differently at the ASGI level:

1. **`uvicorn --root-path /myapp`** — uvicorn prepends `root_path` to `scope["path"]` on every request, so path and root_path are already consistent. Starlette's routing works correctly without any help.

2. **`FastAPI(root_path="/myapp")`** — uvicorn doesn't know about this, so `scope["path"]` is NOT prepended. Path and root_path become inconsistent, and Starlette's nested Mount routing breaks — every admin route returns 404.

The middleware handles case 2. Without it:

```
FastAPI(root_path="/myapp"), middleware REMOVED:
  GET /admin/                    -> 404
  GET /admin/user/list           -> 404
  GET /admin/statics/css/main.css -> 404

FastAPI(root_path="/myapp"), WITH middleware:
  GET /admin/                    -> 200
  GET /admin/user/list           -> 200
  GET /admin/statics/css/main.css -> 200
```

In case 1, the middleware is a no-op — it detects the path is already correct and skips rewriting. So it's safe in both cases.

To your point about separating concerns — the PR does have two independent fixes:

1. **Middleware** — fixes Starlette routing when `root_path` is set via the app constructor
2. **Template `url_for()` for AJAX** — fixes the `data-url` attribute on Select2 widgets to include `root_path`

They solve different problems and could be split into separate PRs if that's preferred, but both are needed for full `root_path` support.

mmzeynalli:
It could be split, but core logic is the same for both fixes, so I plan to keep it as one. @MaximDementyev do you think it is overengineered? Would like to understand why you are not content with @JartanFTW's solution

MaximDementyev:
```
main_app = FastAPI()
api_app = FastAPI()
main_app.mount('/api', api_app)
api_v1_app = FastAPI()
api_app.mount('/v1', api_v1_app)
```
You don't expect an application to be accessible not through the path /api/v1, but directly through /v1. Any sub-application that connects to the main application will be accessible through the main application, and not in any other way.

This logic does not depend on whether an application has a root_path or not. A child always inherits its parent's root_path and adds its own part to it. The idea that a sub-application should be independent of an application is highly unusual. 
It is necessary not to try to change something in order to break the connection with the parent, but to use the tools correctly.

When creating an Admin, an application is created that connects to the FastAPI application. Accordingly, it inherits the root_path from it. This is what we see in the example https://github.com/smithyhq/sqladmin/pull/996#issuecomment-4218709997, where the routing fails because the path is not through the parent. I believe that ignoring the root_path of the parent is incorrect.

Therefore, I believe that middleware does not solve any problems other than correcting incorrect use of the tool.

A few more details about the example:
If there is no nginx in front of the application, using root_path is not the correct approach. 
If the application is behind nginx, then the question arises: should half of the application (the api) be behind nginx, while the other half (the Admin) should bypass nginx and go directly to the application? But is it necessary for this to be a single application? It seems to me that this is a very strange task and infrastructure. To solve this issue, it is necessary to configure nginx correctly.

There may be cases where this strange approach is necessary and there is no way to configure nginx, but I cannot think of any examples.
If such an approach is still necessary, it seems to me that it would be correct to add middleware but not to connect it by default, as authorization is not connected. It will be necessary for an extremely small number of users. They will be able to import it and connect it as they see fit.

JartanFTW:
You're not necessarily wrong, but...

> "Be conservative in what you do, be liberal in what you accept from others." - Postel’s Law

Does the middleware approach break the use-case you're describing?

JartanFTW:
closes #1012

MaximDementyev:
The point of your middleware is to identify requests that are being redirected to a sub-application, bypassing the main application, thus violating the routing and inheritance logic. It then adjusts the path so that the request goes through the main application, thus correcting any previous errors.

JartanFTW:
That is correct, the middleware corrects user errors. I'm a user who made this error.

MaximDementyev:
At the very least, this means that the documentation needs to be updated.

JartanFTW:
Could you give me some more precise direction/requirements on that, please?

mmzeynalli:
I believe we have added more coverage on how the app should be mounted. We did not break anything; we just added more support. Nothing to document here, UNLESS there is a "bypass" that breaks these changes

sqladmin/widgets.py
```diff
@@ -45,7 +45,10 @@ def __init__(self, multiple: bool = False):
 
     def __call__(self, field: "AjaxSelectField", **kwargs: Any) -> Markup:
         kwargs.setdefault("data-role", "select2-ajax")
-        kwargs.setdefault("data-url", field.loader.model_admin.ajax_lookup_url)
+        ajax_url = kwargs.pop("data_url", None)
+        if ajax_url is None:
+            ajax_url = field.loader.model_admin.ajax_lookup_url
+        kwargs.setdefault("data-url", ajax_url)
```
mmzeynalli: I think we can shrink code here:

```python
ajax_url = kwargs.pop("data_url", None) or field.loader.model_admin.ajax_lookup_url
kwargs["data-url"] = ajax_url
```

mmzeynalli (changes requested):
Just to clarify: TestClient(root_path='') simulates ASGI version that we discussed right?

JartanFTW:
TestClient sets scope["root_path"] without prepending to the path, which matches the FastAPI(root_path=...) case, not the uvicorn --root-path case. With modern uvicorn the middleware is a no-op because uvicorn already makes path and root_path consistent. The middleware is needed for the constructor case.

JartanFTW:
Simplified. I realized that WTForms' clean_key normalizes data_url → data-url in kwargs before the widget's \_\_call\_\_ runs. So setdefault("data-url", fallback) already prefers the template-provided value when present. No extra pop logic needed. Retained the leak test in-case WTForms changes.

sqladmin/widgets.py
JartanFTW: https://github.com/smithyhq/sqladmin/pull/996#issuecomment-4232813478

mmzeynalli:
@JartanFTW @MaximDementyev would love if you guys gave your thoughts on my addition:

When I checked locally, with `uvicorn main:app --root-path /api/v1`, I got both for AJAX and statics pathes like

`INFO:     127.0.0.1:49672 - "GET /api/v1/api/v1/admin/statics/js/bootstrap.min.js HTTP/1.1" 304 Not Modified`

So, I added extra if statement (and corresponding test cases) for current PR. Everything else seems to be in place, will merge it soon.

mmzeynalli (approved):

mmzeynalli:
Okay, so after REALLY DEEP testing, I get this result:

We do not need RootPathMiddleware. The problem mentioned at #538 is not relevant, CSS load is not a problem anymore. However, AJAX problem persists.

From my analysis, this is the case for CSS problems (RPM = RootPathMiddleware):

| Case                           | w/ RPM             | wo/ RPM            | w/ RPM + Reverse Proxy               | wo/ RPM + Reverse Proxy         |
| ------------------------------ | ------------------ | ------------------ | ------------------------------------ | ------------------------------- |
| FastAPI(), uvicorn             | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: (no strip)        | :heavy_check_mark: (no strip)   |
| FastAPI(), uvicorn --root-path | :heavy_check_mark: | :x:                | :heavy_check_mark: (does not matter) | :heavy_check_mark: (with strip) |
| FastAPI(root_path=)            | :heavy_check_mark: | :heavy_check_mark:                | :heavy_check_mark: (no strip)        | :heavy_check_mark: (no strip)   |

So, the only case that we need to "cover" is `uvicorn --root-path` command, which we do not need. That command is 99% used with reverse proxy, and we can let user to handle that, no need to cover 1% case. Other than that, everything else works fine. However, we still need to merge this PR because of AJAX url problems. So, @JartanFTW I kindly ask you to restore application.py, remove tests added by me as well

mmzeynalli (changes requested):
As I indicated in previous comment

JartanFTW:
You're right that accessing `/myapp/admin/` directly works without the middleware — I was testing the wrong access pattern.

However, my actual deployment is Traefik stripping the prefix + `FastAPI(root_path=...)` without `--root-path` on uvicorn. FastAPI's docs present this as a [fallback option](https://fastapi.tiangolo.com/advanced/behind-a-proxy/) when the CLI flag isn't available. In that case, requests arrive at `/admin/` (stripped) with `root_path="/myapp"`, and sqladmin 404s without the middleware.

That said, you're the maintainer. If you'd prefer the middleware removed, I'll make the change. Please confirm once more before I proceed.

mmzeynalli:
@JartanFTW can you explain more please? I did `FastAPI(root_path='/api/v1/')` and was able to access admin at `/admin` without `/api/v1`

JartanFTW:
@mmzeynalli Try this and let me know what you get. Maybe I'm overlooking something obvious:

```python
from fastapi import FastAPI
from starlette.testclient import TestClient
from sqladmin.application import RootPathMiddleware

engine = create_engine("sqlite:///test.db")
Base = declarative_base()

class User(Base):
    __tablename__ = "users"
    name = Column(String(32))

app = FastAPI(root_path="/api/v1")
admin = Admin(app, engine)

    pass

admin.add_view(UserAdmin)

# Remove middleware
app.user_middleware = [m for m in app.user_middleware if m.cls is not RootPathMiddleware]
app.middleware_stack = None

with TestClient(app) as client:
    print("GET /admin/ ->", client.get("/admin/").status_code)  # 404
```

mmzeynalli:
/admin is 404, `/api/v1/admin` is working just fine. I think this is just fine, and is more like user behaviour. They can choose to add /api/v1/ to routers instead of whole app. If they choose for whole app, then they mean for admin as well. I think we can remove middleware

JartanFTW:
Here's the full picture with the stripping proxy case added:

| Case | w/ RPM | wo/ RPM | w/ RPM + Reverse Proxy | wo/ RPM + Reverse Proxy |
|------|--------|---------|------------------------|-------------------------|
| `FastAPI()`, uvicorn | ✔️ | ✔️ | ✔️ (no strip) | ✔️ (no strip) |
| `FastAPI()`, `uvicorn --root-path` | ✔️ | ❌ | ✔️ (does not matter) | ✔️ (with strip) |
| `FastAPI(root_path=)` | ✔️ | ✔️ | ✔️ (no strip) | ✔️ (no strip) |
| `FastAPI(root_path=)` + stripping proxy | ✔️ | ❌ | ✔️ (with strip) | ❌ (with strip) |

The last row is my deployment — Traefik strips the prefix, requests arrive at `/admin/`, and without the middleware or `--root-path` it 404s. In my specific case, I can switch to `uvicorn --root-path` to fix it, but without the middleware, we are eliminating the [fastapi documented approach](https://fastapi.tiangolo.com/advanced/behind-a-proxy/) from usage. Just wanted to make the case visible.

I'll make the changes soonish. I have other priorities at the moment.

mmzeynalli:
What's interesting is:

```
Having a proxy with a stripped path prefix, in this case, means that you could declare a path at /app in your code, but then, you add a layer on top (the proxy) that would put your FastAPI application under a path like /api/v1.
```

And we just mount admin on top of fastapi app, which should handle everything that comes with it. 

I will keep this issue in mind, and will research it more. Let's close this for now with AJAX fix, release a new version to fix existing bugs, and come back to this later.

mmzeynalli (approved):

JartanFTW:
@mmzeynalli After trying to fix this in my downstream project implementing SQLAdmin, I realized my last message about my case was inaccurate. My project isn't the last row, it's both of the last two rows. The goal I have is to make it so deployments of my project can strip or not strip at the reverse proxy level as they choose, with the expectation that both cases will work. Switching to uvicorn --root-path does not solve my problem. See my analysis (it's a little ai-y as I've spent way too much time on this at this point, but the point stands):

# `root_path` proxy behavior matrix (sqladmin 0.25.0, no `RootPathMiddleware`)

Test results showing how each `root_path` configuration interacts with the two common reverse-proxy shapes. All four backend configurations were tested against the same application running sqladmin 0.25.0 (which carries the merged changes from PR 996 but not the `RootPathMiddleware`).

## Table 1 — Strip-prefix proxy (proxy removes `/api` before forwarding)

| Backend config | `/api/admin/` | `/api/admin/user/list` | `/api/admin/statics/*.css` | `/api/health` | `/api/docs` |
|---|:---:|:---:|:---:|:---:|:---:|
| **A**: `FastAPI()` + no uvicorn flag | 200 | 200 | 200 | 200 | 200 |
| **B**: `FastAPI()` + `uvicorn --root-path=/api` | 200 | 200 | 200 | 200 | 200 |
| **C**: `FastAPI(root_path="/api")` + no uvicorn flag | **404** | **404** | **404** | 200 | 200 |
| **D**: `FastAPI(root_path="/api")` + `uvicorn --root-path=/api` | 200 | 200 | 200 | 200 | 200 |

## Table 2 — Pass-through proxy (proxy forwards `/api/...` unchanged)

| Backend config | `/api/admin/` | `/api/admin/user/list` | `/api/admin/statics/*.css` | `/api/health` | `/api/docs` |
|---|:---:|:---:|:---:|:---:|:---:|
| **A**: `FastAPI()` + no uvicorn flag | 404 | 404 | 404 | 404 | 404 |
| **B**: `FastAPI()` + `uvicorn --root-path=/api` | 404 | 404 | 404 | 404 | 404 |
| **C**: `FastAPI(root_path="/api")` + no uvicorn flag | 200 | 200 | 200 | 200 | 200 |
| **D**: `FastAPI(root_path="/api")` + `uvicorn --root-path=/api` | 404 | 404 | 404 | 404 | 404 |

## Key observations

1. **No single configuration works under both proxy shapes** without `RootPathMiddleware`. Each of the four configurations is correct under exactly one of the two proxy behaviors — there is no overlap. Config B/D cover strip-prefix; Config C covers pass-through; Config A incidentally works under strip-prefix but cannot emit correct absolute URLs without `root_path` configured somewhere.

2. **Config C's strip-prefix row is the exact bug the removed `RootPathMiddleware` was addressing.** Top-level routes (`/health`, `/docs`) return 200, but every nested-Mount route (admin pages *and* admin static files) returns 404. The cause: with `FastAPI(root_path="/api")` and a strip-prefix proxy, requests arrive with `scope[path]=/admin/` and `scope[root_path]=/api`. Starlette's `get_route_path()` skips stripping because `/admin/` does not start with `/api`, so top-level routing at `/admin` still matches — but the nested `Mount("/admin", ...)` then delegates into sqladmin's inner Starlette app, whose own routing re-derives its prefix and fails to match any child route.

3. **Config D ("belt-and-suspenders") does not cover both cases.** Setting both the constructor arg and the uvicorn flag simply degrades to the same behavior as Config B. This is because uvicorn prepends `root_path` to `scope[path]` before FastAPI's `__call__` runs, so Starlette's subsequent single strip leaves identical final state regardless of whether the constructor arg was also set.

4. **Config A's strip-prefix row is green only incidentally.** With no `root_path` configured at either layer, routes at `/admin/`, `/health/`, etc. match because the strip-prefix proxy delivers those paths verbatim. But the FastAPI app has `root_path=""`, so any absolute URL the framework builds (OpenAPI `servers` entry, `request.url_for()` output in admin/docs responses) drops the `/api` prefix — which breaks the browser-facing flow even when pure routing works. Not a production-usable configuration.

## Implication for deployment choice

With sqladmin 0.25.0, the choice is proxy-shape-bound:

- **Deploy with a strip-prefix proxy (Traefik `StripPrefix`, nginx `rewrite`, Vite/Next `rewrites`)** → use `uvicorn --root-path=/api`, leave the constructor arg unset. Matches Config B.
- **Deploy with a pass-through proxy** → use `FastAPI(root_path="/api")`, leave the uvicorn flag unset. Matches Config C. (Including admin static files — this PR 996 template fix for `url_for()` is what makes `/admin/statics/*.css` return 200 in Table 2 Row C; prior to 0.25.0 it returned 404.)
- **Do not set both** — no benefit, and Config D's behavior can mislead operators into thinking they have coverage they do not have.

## Reproduction

Tested on FastAPI 0.127, Starlette 0.50, sqladmin 0.25.0, uvicorn 0.40 (httptools). SQLAdmin is mounted at `/admin` as a nested `starlette.Mount` on a single FastAPI app. Strip-prefix proxy is the Vite dev proxy (`rewrite: (path) => path.replace(/^\/api/, "")`); pass-through proxy is a minimal nginx container with `location /api/ { proxy_pass http://backend:8000; }` (no URI rewrite). Each config was exercised by toggling the `FastAPI(root_path=...)` constructor arg and the `uvicorn --root-path` CLI flag, rebuilding the container between variants. Probes were curled from a sibling container on the same Docker network, carrying a superuser session cookie obtained from my project's `POST /api/auth/login` against each proxy (with `COOKIE_SECURE=false` so curl retains the `Secure` cookie over plain HTTP).

## What does this mean?

This means there is no `root_path` configuration on sqladmin 0.25.0 that lets a single deployment accept either a strip-prefix or a pass-through proxy. Because the config lives in code, not runtime env, I have to hard-code a proxy shape for the whole project — every operator inherits that choice, and any operator whose infrastructure uses the other shape silently loses every admin route. `RootPathMiddleware` was what closed that gap — with it present, a single backend config covered both rows of both tables. Would you reconsider reintroducing it (even gated behind an opt-in import)? That would restore deployment flexibility for downstream projects that can't constrain the proxy shape their operators choose.

MaximDementyev:
@JartanFTW  It seems to me that it would be correct to compare your tests with the usual 2 fastapi applications to understand whether we are trying to fix bugs in sqladmin or fastapi/starlette. For example:
```
app = FastAPI(root_path='/api')
app2 = FastAPI()
app.mount(app=app2, path='/admin')
@app2.get('/test')
def test():
    return True
```

> Because the config lives in code, not runtime env, I have to hard-code a proxy shape for the whole project — every operator inherits that choice, and any operator whose infrastructure uses the other shape silently loses every admin route.

What do you mean? 
root-path uvicorn - is set in the runtime environment
root-path fastapi - the developer can add it to the application settings, and it will be set at the runtime  environment

JartanFTW:
Mmm you have a point, this would be an upstream concern now.

To clarify my earlier point anyway: You're right that the value can come from env vars. But the mechanism "use constructor arg" vs "rely on --root-path" is a code decision that typically occurs at different levels, making it much more involved to allow for either/or. The value is easily configurable, but not the mechanism per deployment, which is what's limiting downstream flexibility.

JartanFTW:
Mentioned in the relevant upstream fastapi discussion: https://github.com/fastapi/fastapi/discussions/15132#discussioncomment-16705219

Edit: They requested a dedicated discussion post: https://github.com/fastapi/fastapi/discussions/15430

PR #999, opened 2026-02-06
[Feature] Add `column_type_formatters_detail`

maxim-f1:
# A new `column_type_formatters_detail` field has been added to `ModelView`.

1. This field follows the same principle as `column_type_formatters`, with the sole difference that it is only applied on the details page.

2. If `column_type_formatters_detail` is not set, the details page will fall back to using formatting from `column_type_formatters`.

3. Several additional built-in formatters have been introduced (the default `column_type_formatters` remains unchanged).

4. Now, when a parent class is used as the key for `column_type_formatters` or `column_type_formatters_detail`, all its child classes will also be formatted accordingly. 

5. If both a child and its parent are specified, the child's formatter takes precedence.

# New formatters

### 1. `StrEnum` Formatter
Adds a tooltip displaying the list of available enum values.

#### List:
str_enum_title_list

#### Details:
str_enum_title_details

### 2. Copy String Formatter
Adds a copy-to-clipboard button for any string field content.

#### List:
copy_list

#### Details:
copy_details_1

copy_details_2

### 3. `date` or `datetime` Formatter
Adds a tooltip with the precise timestamp and displays the date in a human-readable format.

#### List:
datetime_list

#### Details:
datetime_details


mmzeynalli:
@maxim-f1 any updates on this one? Shall we keep it open?

maxim-f1:
@mmzeynalli 

Will doing everything as I want be harmful because Python <3.11 doesn't support StrEnum. My options are:

1) I can remove all uses of StrEnum, but then I'll have to remove the str_enum_formatter formatting template.

2) I'll try to support this functionality in older versions.

I'm leaning toward the first option.

sqladmin/models.py
```diff
@@ -63,6 +66,16 @@
...
+# Import StrEnum for python < 3.10 and > 3.11
+if sys.version_info < (3, 11):
+
+    class StrEnum(str, Enum):
+        __str__ = str.__str__
+        __repr__ = Enum.__repr__
+else:
+    from enum import StrEnum as StrEnum  # noqa: F401
+
+
```
mmzeynalli: import StrEnum from formatters

mmzeynalli:
This is nice feature, if possible finalize this. I think you can keep StrEnum with the hack you used in formatters.py.

mmzeynalli (approved):

PR #1004, opened 2026-02-27
Added default_value to Filters

wnowicki:
### Added

- Default value to the `BooleanFilter` and `StaticValuesFilter`

### Fixed 

- Issue: [B704:markupsafe_markup_xss] Potential XSS with ``markupsafe.Markup`` detected. Do not use ``Markup`` on untrusted data.


aminalaee (changes requested):
Thanks for the PR. Can you please:
- Add tests and update description of PR
- Remove changes uv.lock (not needed anymore)
- Either make default_value optional in the Protocol (with a default) or add it to all filter classes

mmzeynalli:
@wnowicki, what would happen if I have a nullable bool field in the database? Let's say, status:

`None` means pending
`False` means rejected
`True` means accepted?

In this case, if I want `None` to be the default value, it won't work properly.

Instead, I suggest this:

```python
T = TypeVar('T')

_UNSET = object()

Unset = Union[T, Literal[UNSET]]
UnsetN = Union[T, Literal[UNSET], None]

defaul_value: UnsetN = _UNSET
```

And other changes accordingly. Now, None and unset default are two different values.

mmzeynalli (changes requested):
Also believe, it wont be hard to add default logic to `AllUniqueStringValuesFilter` and `StaticValuesFilter`. Would love to that too

wnowicki:
> @wnowicki, what would happen if I have a nullable bool field in the database? Let's say, status:
> 
> `None` means pending `False` means rejected `True` means accepted?
> 
> In this case, if I want `None` to be the default value, it won't work properly.
> 
> Instead, I suggest this:
> 
> ```python
> T = TypeVar('T')
> 
> _UNSET = object()
> 
> Unset = Union[T, Literal[UNSET]]
> UnsetN = Union[T, Literal[UNSET], None]
> 
> defaul_value: UnsetN = _UNSET
> ```
> 
> And other changes accordingly. Now, None and unset default are two different values.

@mmzeynalli I was a little puzzled with that proposition from the beginning, although I can see some limited usage to `NULL` values, case you gave as na example would be better solved by setting explicit statuses instead of using `bool`

I've implemented it and made it working, we know how flexible Python is. But when it comes to committing it cannot pass checks.

`Literal` must have string as the first argument not an object, then if we do `Literal["_UNSET"]` the actual literal is expected when it comes to setting it up.

```python
default_value: UnsetAny = "_UNSET",
```

Please review changes in the next commit but treat it more as work in progress, I'm not sure if this is the right direction.

Regarding default values on other filters, I'd suggest to get this scope done first.

mmzeynalli:
Seems okay. Although, I cannot agree with part:

`Literal must have string as the first argument not an object,`, because I have used this logic in different library.

Other than that, I think it is good to have that None and Unset differentiation now, because, even its use for bool is limited as you mentioned, I do not think it is limited for other fields.

Please fix Unset to be object, and fix merge conflicts

mmzeynalli (approved):

wnowicki:
Thanks @mmzeynalli. I didn't have time recently to complete it.

PR #1027, opened 2026-04-07
Implementation multiple admin panels

### LINKED ISSUE #738: Multiple Admin Instances result in wrong URL's to the views
### Checklist

### Describe the bug

I have an application with the option of extensions. My idea was to have an admin page for the main application and separate admin pages for the extensions.

You can have multiple instances of `Admin` running, but the links to the Views on the left point to wrong URL's.

1. Start the app
2. The Views are working correctly, when accessing with the url

    1. http://localhost:8000/admin/user1/list 
    2. http://localhost:8000/sub-app/admin/user2/list 
  
4. The navigation results in wrong url
    1. http://localhost:8000/admin --> Click on User1s -> Wrong URL `http://localhost:8001/sub-app/admin/user1/list`

```python

from fastapi import FastAPI


# DB -------------------------------------------------------------------------
Base = declarative_base()
engine = create_engine(
    "sqlite:///example.db",
)


class User1(Base):
    __tablename__ = "users1"

    name = Column(String)


class User2(Base):
    __tablename__ = "users2"

    name = Column(String)


Base.metadata.create_all(engine)  # Create tables

# Main App -------------------------------------------------------------------
app = FastAPI()

# mounted Sub-App ------------------------------------------------------------
sub_app = FastAPI()
app.mount('/sub-app', sub_app)

# Admin ----------------------------------------------------------------------
admin1 = Admin(app, engine, base_url="/admin")
admin2 = Admin(sub_app, engine, base_url="/admin")


class UserAdmin1(ModelView, model=User1):
    column_list = [User1.id, User1.name]


class UserAdmin2(ModelView, model=User2):
    column_list = [User2.id, User2.name]


admin1.add_view(UserAdmin1)
admin2.add_view(UserAdmin2)

```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

Tested on Windows, I could also test on Linux

- Windows 10
- Python: `3.10.2`
- Sqladmin: `0.16.1`

### Additional context

_No response_

### LINKED ISSUE #919: Multiple Admin Panels for the same fastapi App
### Checklist

I want to create multiple admin panels for one fastapi app.

Adjust all the code to use dynamic "mount_name" instead of fixed names. 

_No response_

### Additional context

```
class CustomAdmin(Admin):
    def __init__(
        self,
        app: Starlette,
        engine: ENGINE_TYPE | None = None,
        session_maker: sessionmaker | async_sessionmaker | None = None,
        base_url: str = "/admin",
        title: str = "Admin",
        logo_url: str | None = None,
        favicon_url: str | None = None,
        middlewares: Sequence[Middleware] | None = None,
        debug: bool = False,
        authentication_backend: AuthenticationBackend | None = None,
        mount_name: str = "admin",
    ) -> None:
        """
        Args:
            app: Starlette or FastAPI application.
            engine: SQLAlchemy engine instance.
            session_maker: SQLAlchemy sessionmaker instance.
            base_url: Base URL for Admin interface.
            title: Admin title.
            logo_url: URL of logo to be displayed instead of title.
            favicon_url: URL of favicon to be displayed.
        """

        super().__init__(
            app=app,
            engine=engine,
            session_maker=session_maker,
            base_url=base_url,
            title=title,
            logo_url=logo_url,
            favicon_url=favicon_url,
            templates_dir=templates_dir,
            middlewares=middlewares,
            authentication_backend=authentication_backend,
        )

        statics = StaticFiles(packages=["sqladmin"])

        async def http_exception(
            request: Request, exc: Exception
        ) -> Response | Awaitable[Response]:
            assert isinstance(exc, HTTPException)
            context = {
                "status_code": exc.status_code,
                "message": exc.detail,
            }
            return await self.templates.TemplateResponse(
                request, "sqladmin/error.html", context, status_code=exc.status_code
            )

        routes = [
            Mount("/statics", app=statics, name=f"{mount_name}_statics"),
            Route("/", endpoint=self.index, name=f"{mount_name}_index"),
            Route("/{identity}/list", endpoint=self.list, name=f"{mount_name}_list"),
            Route(
                "/{identity}/details/{pk:path}", endpoint=self.details, name=f"{mount_name}_details"
            ),
            Route(
                "/{identity}/delete",
                endpoint=self.delete,
                name=f"{mount_name}_delete",
                methods=["DELETE"],
            ),
            Route(
                "/{identity}/create",
                endpoint=self.create,
                name=f"{mount_name}_create",
                methods=["GET", "POST"],
            ),
            Route(
                "/{identity}/edit/{pk:path}",
                endpoint=self.edit,
                name=f"{mount_name}_edit",
                methods=["GET", "POST"],
            ),
            Route(
                "/{identity}/export/{export_type}", endpoint=self.export, name=f"{mount_name}_export"
            ),
            Route(
                "/{identity}/ajax/lookup", endpoint=self.ajax_lookup, name=f"{mount_name}_ajax_lookup"
            ),
            Route("/login", endpoint=self.login, name=f"{mount_name}_login", methods=["GET", "POST"]),
            Route("/logout", endpoint=self.logout, name=f"{mount_name}_logout", methods=["GET"]),
        ]

        self.admin.router.routes = routes
        self.admin.exception_handlers = {HTTPException: http_exception}
        self.admin.debug = debug
        self.app.mount(base_url, app=self.admin, name=mount_name)
```

MaximDementyev:
### Problem
It is not possible to add multiple administration panels within the same application

1. SqlAdmin connects to the FastApi with a fixed name Since query routing is calculated by name, no matter how many Admin applications we create, all links will lead to the first application created. Creating an additional name parameter is not enough, it is necessary to rewrite all url calculations, request.url_for("admin:list") is used everywhere.
2. The ModelView class uses the session_maker variable, which is a class variable. Therefore, when adding a second application, SqlAdmin session_maker will be overwritten. It turns out that no matter how many SqlAdmin applications we add, they will all work with the database added last.
fix #919 

### Solution

1. In sqladmin.helpers two methods have been created: 
**local_url_for** - Generate a URL for the specified route, taking router hierarchy into account.
**get_current_router_name** - Find the router name in the hierarchy that corresponds to the target router. Traverses the routes of the start router and its nested applications (of type `Mount`) to find a match with the target router.
The entire project has updated the use of request.url_for to local_url_for.
2. session_maker is now an attribute of the class object

This made it possible to run multiple sqladmins within the same application, and also remove the restriction that SqlAdmin can only be connected to the FastApi root application.
```
from fastapi import FastAPI
from sqladmin import action

Base = declarative_base()
engine1 = create_engine("sqlite:///example1.db",connect_args={"check_same_thread": False})
engine2 = create_engine("sqlite:///example2.db",connect_args={"check_same_thread": False})
engine3 = create_engine("sqlite:///example3.db",connect_args={"check_same_thread": False})

class User(Base):
    __tablename__ = "users"
    name = Column(String)

Base.metadata.create_all(engine1)
Base.metadata.create_all(engine2)
Base.metadata.create_all(engine3)

app = FastAPI()
admin1 = Admin(app=app, engine=engine1, base_url='/admin/base1', mount_name='base1')
admin1.add_view(UserAdmin)

admin2 = Admin(app=app, engine=engine2, base_url='/admin/base2', mount_name='base2')
admin2.add_view(UserAdmin)

app3 = FastAPI()
app.mount('/api', app3, 'api')
admin3 = Admin(app=app3, engine=engine3, base_url='/admin/base3', mount_name='base3')
admin3.add_view(UserAdmin)
```
In the example, three administration panels are connected that are connected to different databases
admin1 is accessible via the path /admin/base1
admin2 is accessible via the path /admin/base2
admin3 is accessible via the path /api/admin/base3

If this solution is suitable, I will update the documentation and add tests for new functions.

### DIFF
--- a/sqladmin/_menu.py
+++ b/sqladmin/_menu.py
@@ -8,6 +8,8 @@ from starlette.requests import Request
 if TYPE_CHECKING:
     from sqladmin.application import BaseView, ModelView
 
+from sqladmin.helpers import local_url_for
+
 
 class ItemMenu:
     def __init__(self, name: str, icon: str | None = None) -> None:
@@ -73,8 +75,8 @@ class ViewMenu(ItemMenu):
 
     def url(self, request: Request) -> str | URL:
         if self.view.is_model:
-            return request.url_for("admin:list", identity=self.view.identity)
-        return request.url_for(f"admin:{self.view.identity}")
+            return local_url_for(request, "list", identity=self.view.identity)
+        return local_url_for(request, self.view.identity)
 
     @property
     def display_name(self) -> str:
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -41,6 +41,7 @@ from sqladmin.forms import WTFORMS_ATTRS, WTFORMS_ATTRS_REVERSED
 from sqladmin.helpers import (
     get_object_identifier,
     is_async_session_maker,
+    local_url_for,
     slugify_action_name,
 )
 from sqladmin.models import BaseView, ModelView
@@ -236,13 +237,12 @@ class BaseAdmin:
 
         view._admin_ref = self
         # Set database engine from Admin instance
-        view.session_maker = self.session_maker
         view.is_async = self.is_async
         view.ajax_lookup_url = urljoin(
             self.base_url + "/", f"{view.identity}/ajax/lookup"
         )
         view.templates = self.templates
-        view_instance = view()
+        view_instance = view(self.session_maker)
 
         self._find_decorated_funcs(
             view, view_instance, self._handle_action_decorated_func
@@ -367,6 +367,7 @@ class Admin(BaseAdminView):
         debug: bool = False,
         templates_dir: str = "templates",
         authentication_backend: AuthenticationBackend | None = None,
+        mount_name: str = "admin",
     ) -> None:
         """
         Args:
@@ -446,7 +447,7 @@ class Admin(BaseAdminView):
         self.admin.router.routes = routes
         self.admin.exception_handlers = {HTTPException: http_exception}
         self.admin.debug = debug
-        self.app.mount(base_url, app=self.admin, name="admin")
+        self.app.mount(base_url, app=self.admin, name=mount_name)
 
     @login_required
     async def index(self, request: Request) -> Response:
@@ -519,7 +520,7 @@ class Admin(BaseAdminView):
 
         referer_url = URL(request.headers.get("referer", ""))
         referer_params = MultiDict(parse_qsl(referer_url.query))
-        url = URL(str(request.url_for("admin:list", identity=identity)))
+        url = URL(str(local_url_for(request, "list", identity=identity)))
         url = url.include_query_params(**referer_params)
         return PlainTextResponse(content=str(url))
 
@@ -658,7 +659,7 @@ class Admin(BaseAdminView):
                 request, "sqladmin/login.html", context, status_code=400
             )
 
-        return RedirectResponse(request.url_for("admin:index"), status_code=302)
+        return RedirectResponse(local_url_for(request, "index"), status_code=302)
 
     async def logout(self, request: Request) -> Response:
         if self.authentication_backend is None:
@@ -672,7 +673,7 @@ class Admin(BaseAdminView):
         if isinstance(response, Response):
             return response
 
-        return RedirectResponse(request.url_for("admin:index"), status_code=302)
+        return RedirectResponse(local_url_for(request, "index"), status_code=302)
 
     async def ajax_lookup(self, request: Request) -> Response:
         """Ajax lookup route."""
@@ -706,14 +707,14 @@ class Admin(BaseAdminView):
         identifier = get_object_identifier(obj)
 
         if form.get("save") == "Save":
-            return request.url_for("admin:list", identity=identity)
+            return local_url_for(request, "list", identity=identity)
 
         if form.get("save") == "Save and continue editing" or (
             form.get("save") == "Save as new" and model_view.save_as_continue
         ):
-            return request.url_for("admin:edit", identity=identity, pk=identifier)
+            return local_url_for(request, "edit", identity=identity, pk=identifier)
 
-        return request.url_for("admin:create", identity=identity)
+        return local_url_for(request, "create", identity=identity)
 
     async def _handle_form_data(self, request: Request, obj: Any = None) -> FormData:
         """
--- a/sqladmin/authentication.py
+++ b/sqladmin/authentication.py
@@ -8,6 +8,8 @@ from starlette.middleware import Middleware
 from starlette.requests import Request
 from starlette.responses import RedirectResponse, Response
 
+from sqladmin.helpers import local_url_for
+
 
 class AuthenticationBackend:
     """Base class for implementing the Authentication into SQLAdmin.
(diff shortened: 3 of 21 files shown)


mmzeynalli:
@MaximDementyev right now, im reviewing smaller PRs, I need enough time for focus to review this, so it can take time. I have not forgotten about this one)

MaximDementyev:
The [FastApi 0.137.0](https://fastapi.tiangolo.com/release-notes/#01370-2026-06-14) update has broken the routers.routers traversal logic.
It will take some time to figure out how to fix this.

If implementing any url search logic seems complicated, you can consider creating a function like this:
```
def local_url_for(request, name, **kwargs):
   return request.url_for(f"admin:{name}", **kwargs)
```

This will allow you to collect all the URL search logic in one place, giving users the ability to override this function. 
It is very difficult to redefine url_for now, as it affects many files.

mmzeynalli:
This is incomplete. I am putting this to draft for now. Here are things that should be considered/fixed

1. Document breaking changes (especially moving from classvar to object var, i.e. `session_maker`
2. The other classvars are still being overriden: 

```
view._admin_ref = self
view.is_async = self.is_async
view.ajax_lookup_url = urljoin(
        self.base_url + "/", f"{view.identity}/ajax/lookup"
)
view.templates = self.templates
view_instance = view(self.session_maker)
```

So if two admin have different ajax_lookup_urls, it will be messed up

3. For backwards compatibility, user overriden templates should work. As of today, if user overrode template, it will be `admin:admin:list`.

4. Document mount_name param, both in markdown and in the code
5. Add tests that have two admins
6. `docs/authentication.md` and `docs/configurations.md` still tell users to write request.url_for("admin:index") 
7. If there is unnamed mounts, in `get_current_router_name`, a nested match returns f"{router.name}:{sub_name}" — if the intermediate Mount has no name, that's the literal string "None:admin", and URL generation blows up.
8.  Dead comparison in the first loop. for router in `getattr(start_router, "routes", []): if router == target_router` compares a `Route/Mount` against a `Router` — never true.
9. O(route-tree) walk per generated URL. A list page emits dozens of URLs (per-row view/edit/delete, pagination, exports, statics ×13); each one recursively walks the host app's entire route tree. On a large FastAPI app that's thousands of comparisons per request for a value that never changes. The prefix is static — compute it once at mount time (the Admin literally receives mount_name as a parameter!) or cache it in request.scope on first use.

OR:

Inside an admin request, `request.app` is the admin's Starlette app, so `request.app.url_path_for(name, **params).make_absolute_url(request.base_url)` produces the correct URL with no name-prefix discovery, no `mount_name` uniqueness requirement, no tree walk

MaximDementyev:
@mmzeynalli
2\. The problem is only with ajax_lookup_url. The rest is correct, as it is overridden from the Admin object.
With ajax, it’s indeed more complicated. Since the URL is static and is calculated in advance (when we don’t have a request), it’s not possible to determine the full mounting path without additional information.

3\. What’s the best way to fix it? I can avoid changing the url_for function in the templates and instead add a new one — then all the old code will still work.

7\. If there is an application mounted without a name in the path, we won’t be able to use request.url_for. When searching for a URL by name, there should be a restriction to non‑empty names. We can explicitly check this and raise our own error, or we can describe this point in the documentation.

9\. Мы не можем вычислить префикс самостоятельно, так как он будет относиться к родительскому элементу.  Чтобы вычислить полный префикс, нам нужен либо запрос, либо явное указание пользователя передать его нам.  Кэширование в области видимости позволит нам не вычислять путь для static.  (один вызов для static и требуемый путь) Возможно, лучшим решением для кэширования будет добавление переменной в приложение Starlette (self.admin, которая монтируется) и изменение ее при первом запросе.

request.app.url_path_for(name, **params).make_absolute_url(request.base_url) не работает. Поскольку base_url не содержит всех префиксов подключенных приложений. В приведенном выше примере он будет равен `http://127.0.0.1:8000`, а пути будут соответствовать аналогичному URL-адресу `http://127.0.0.1:8000/statics/css/tabler.min.css`, что неверно.

It’s still unclear what to do about updating fastapi and the prohibition on iterating over routes.
I can see one option for now. That is to ask the user for the full mount prefix. Then there will be no problems with ajax, and the full path can be found via request.app.url_path_for(name, **kwargs).make_absolute_url(request.base_url + full_prefix)

PR #1035, opened 2026-04-18
fix: authenticate ajax lookup endpoint

### LINKED ISSUE #1011: AJAX endpoint is not authenticated
### Checklist

### Describe the bug

The AJAX lookup endpoint (`/{identity}/ajax/lookup`) does not require any kind of authentication. This opens the possibility for anyone to view your data with a request URL like:
```
http://localhost:8000/admin/membership/ajax/lookup?name=user&term=test
```

As a workaround, I wrapped the endpoint with `login_required`. Of course, it's not the proper fix because this decorator is intended for HTML requests.
```python
from sqladmin import Admin
from sqladmin.authentication import login_required

class CustomAdmin(Admin):
    # adds missing authentication to the ajax lookup endpoint
    @login_required
    async def ajax_lookup(self, *args, **kwargs):
        return await super().ajax_lookup(*args, **kwargs)
```

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

MacOS, Python 3.13. SQLAdmin 0.23.0

### Additional context

_No response_

vahidzhe:
Fixes #1011

The `/{identity}/ajax/lookup` endpoint was publicly accessible without any authentication, allowing unauthenticated users to query model data directly via a URL like:

`/admin/membership/ajax/lookup?name=user&term=test`

## Changes

**`sqladmin/application.py`**
- Added authentication check at the start of `ajax_lookup()`, consistent with how other protected endpoints are handled. Unauthenticated requests are redirected to the login page.

**`tests/test_authentication.py`**
- Added 3 tests covering unauthenticated access, authenticated access, and access after logout.

### DIFF
--- a/tests/test_authentication.py
+++ b/tests/test_authentication.py
@@ -1,9 +1,9 @@
 from typing import Generator
 
 import pytest
-from sqlalchemy import Column, Integer
+from sqlalchemy import Column, ForeignKey, Integer, String
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
-from sqlalchemy.orm import declarative_base
+from sqlalchemy.orm import declarative_base, relationship
 from starlette.applications import Starlette
 from starlette.requests import Request
 from starlette.responses import JSONResponse, RedirectResponse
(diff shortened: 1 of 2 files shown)


tests/test_authentication.py
```diff
@@ -118,3 +118,82 @@ def test_action_access_login_required_views(client: TestClient) -> None:
 
     response = client.get("/admin/movie/action/test")
     assert {"status": "ok"} == response.json()
+
+
+class Artist(Base):
+    __tablename__ = "artists_auth"
+
+    id = Column(Integer, primary_key=True)
+    name = Column(String(50))
```
aminalaee, line 127: Do we need to add these models and boilerplate to test this? Can't the existing model do the same?

aminalaee (commented):
Implementation looks good to me 👍

tests/test_authentication.py
vahidzhe: The unauthenticated and post-logout tests can use Movie, but the authenticated test needs a model with form_ajax_refs to verify that the endpoint actually returns results (not just that it doesn't redirect). I've kept Artist and SongAuth only for that purpose, and they reuse the existing admin and client - no separate app

aminalaee (approved):
Thanks @vahidzhe for the contribution 👍

PR #1039, opened 2026-04-29
[Fix] AJAX fields: persist selected values after validation errors

maxim-f1:
## How to reproduce:
1. Select a new AJAX value for the field.
2. Trigger a validation error (e.g., by submitting the form with other invalid data).
3. The selected values roll back to their original entity state.

## How it should be:
Selected AJAX field values ​​are saved in step 3.

### Before:
Screenshot from 2026-04-29 17-41-33
Screenshot from 2026-04-29 17-41-44

### After:
Screenshot from 2026-04-29 17-42-25
Screenshot from 2026-04-29 17-42-34

## Detail changes:
* Add `format_by_pk`
* Fix `form` in context after form validation
* Now, even with override Field, the loader field will be used (for example, for custom AJAX widgets)
* If the loader field is None, an empty list will be returned instead of an error.
* If the model has not been transferred, it will be automatically loaded by the PK field.


mmzeynalli:
Interesting. I believe, going back to old value with error makes more sense, as it can be confusing to user to select "wrong" value, and UI to enable them to choose it. But still, I would like to get @aminalaee's opinion as well.

Also, could you explain what format_by_pk is?

maxim-f1:
1. If we believe that input data should roll back when an error occurs, then during creation, the data shouldn't be returned either. (This is currently not the case.)

image



2. By default, an AJAX field enters the widget as a SQLAlchemy model instance. However, if the widget is rendered during an error stage, the AJAX field becomes a PK and won't function correctly. `format_by_pk` is needed for those cases where a PK is passed to the widget instead of a model.

3. Regarding user convenience: Imagine you need to select 50 users in this field, but an error occurs in a completely different field. The AJAX field is perfectly valid, but it will still be rolled back.

aminalaee:
@mmzeynalli I think keeping the user selected value is a nicer experience, but about your concern, can't the user clear/unset the selected ajax value?

sqladmin/ajax.py
```diff
@@ -61,6 +61,28 @@ def format(self, model: type) -> dict[str, Any]:
...
+        filters = [field == pk for field in inspect(self.model).primary_key]
+
+        stmt = stmt.filter(and_(*filters))
+
+        if self.order_by:
+            if isinstance(self.order_by, list):
+                for o in self.order_by:
+                    stmt = stmt.order_by(o)
+            else:
+                stmt = stmt.order_by(self.order_by)
```
mmzeynalli, line 91: Considering PK is unique, why do we need order_by? Also limit(1)?

sqladmin/ajax.py
```diff
@@ -61,6 +61,28 @@ def format(self, model: type) -> dict[str, Any]:
 
         return {"id": str(get_object_identifier(model)), "text": str(model)}
 
+    async def format_by_pk(self, pk: Any) -> dict[str, Any]:
+        stmt = select(self.model)
+
+        filters = [field == pk for field in inspect(self.model).primary_key]
+
```
mmzeynalli: What is value of pk, in case of multi-column PK?

mmzeynalli (changes requested):

mmzeynalli:
@maxim-f1 Any updates?

mmzeynalli (approved):

Issue #1000, opened 2026-02-07
Speed slowdown issue while opening edit page in model


BrikozO:
### Checklist

- [x] The bug is reproducible against the latest release or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Hello! I am using sqladmin and there is a problem: when i am creating model which linked with other model (many to many, many to one, doesn't matter) and create more than 1000 objects in DB, speed of page loading extremely go down.

### Steps to reproduce the bug

For example, i have table users:

```py
class User(SQLAlchemyBaseUserTableUUID, Base):
    roles: Mapped[list['Role']] = relationship(secondary='user_role', back_populates='users')
```

Roles and user_role (many to many):
```py
class Role(Base):
    name: Mapped[str] = mapped_column(unique=True, nullable=False)

    users: Mapped[list['User']] = relationship(secondary='user_role', back_populates='roles')

    def __str__(self):
        return self.name


user_role = Table(
    'user_role',
    EmptyBase.metadata,
    Column('user_id', ForeignKey('user.id'), primary_key=True),
    Column('role_id', ForeignKey('role.id'), primary_key=True),
)
```

I create 3000 roles to test it and link with one user. Try to open page and then - it loads very very slow :(

### Environment

Python 3.12.12

sqladmin 0.21.0 (i am trying to use 0.23 but there is a problem with uuid, i don't think that there is a big difference for my case)

### Additional context

Is there any options to optimize some code inside library? I saw same question in another issue: https://github.com/aminalaee/sqladmin/issues/704

But it's very old, and there are no any status update yet

mmzeynalli:
Can you please share your admin model code?

BrikozO:
```py
class UserAdmin(ModelView, model=User):
    column_list = [User.id, User.email]
    page_size = 50

    form_ajax_refs = {
        "roles": {
            "fields": ("name",),
            "order_by": "id",
            "page_size": 20,
        }
    }
```

I am trying to provide page size param (claude says that shall helps :)) but nothing changes

Oh, i think i need to provide more details - i need customization on this page (add, delete roles, etc) and provide ajax refs param for this

arturbent0:
Hi, i would like to help with this issue if that's okay @mmzeynalli

mmzeynalli:
@arturbent0 sure, go ahead, it is all yours. Do please let us know if you abandon working on it, so we work on it ourselves. Thanks!

[...]

PR #1042, opened 2026-05-04
[Fix] AJAX Select2 field. Removed separate fields for each selection.

### LINKED ISSUE #835: Edit database 400 error: Too many fields. Maximum number of fields is 1000.
### Checklist

### Describe the bug

I have an object that I am trying to edit from the admin panel. The model field is a list of enum values for which I want to add one of the values to the list. The annoying thing is that for some objects with the same model I can alter it fine. However, for some objects the admin page shows the message:
image

I know this is not much to go on but does anyone have any idea what it might be and how I can further my investigation? Please let me know if you need more info

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

- The code is deployed with FastAPI using Dockerfile to a kubernetes cluster
- The docker base image is python:3.12.3-slim https://hub.docker.com/layers/library/python/3.12.3-slim/images/sha256-a9922d9d812c99034d23bef370270abe95def980c1f2664afac002f3e6a9428e

### Additional context

_No response_

maxim-f1:
# https://github.com/smithyhq/sqladmin/issues/1000
# Fix "Too many fields" error for many selection in AJAX

## Before:
image

## After:
Screenshot from 2026-05-04 11-48-18

# Details:
* Instead of separate fields for each selection, all AJAX values are passed as a single comma-separated field.
* Added a placeholder.
* Added a "Clear All" button to remove all inputs.

### DIFF
--- a/sqladmin/statics/js/main.js
+++ b/sqladmin/statics/js/main.js
@@ -98,28 +98,44 @@ $(':input[data-role="datetimepicker"]:not([readonly])').each(function () {
 
 // Ajax Refs
 $(':input[data-role="select2-ajax"]').each(function () {
-  $(this).select2({
+  var placeholderText = $(this).attr("placeholder") || "Search";
+
+  var $select = $(this);
+  var originalName = $select.attr("name");
+  var $hiddenInput = $('<input type="hidden">').attr("name", originalName);
+  $select.after($hiddenInput);
+  $select.removeAttr("name");
+
+  $select.select2({
     minimumInputLength: 1,
+    placeholder: placeholderText,
+    allowClear: true,
     ajax: {
-      url: $(this).data("url"),
+      url: $select.data("url"),
       dataType: 'json',
       data: function (params) {
-        var query = {
-          name: $(this).attr("name"),
+        return {
+          name: originalName,
           term: params.term,
-        }
-        return query;
+        };
       }
     }
   });
 
-  existing_data = $(this).data("json") || [];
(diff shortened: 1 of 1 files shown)


sqladmin/statics/js/main.js
```diff
@@ -98,29 +98,45 @@ $(':input[data-role="datetimepicker"]:not([readonly])').each(function () {
...
     }
   });
 
-  existing_data = $(this).data("json") || [];
-  for (var i = 0; i < existing_data.length; i++) {
-    data = existing_data[i];
+  $select.on('change', function () {
+    var val = $select.val();
+    $hiddenInput.val(val ? val.join(',') : '');
+  });
```
mmzeynalli, line 128: It should be something like this:

```js
  $select.on('change', function () {
    var val = $select.val();
    if (Array.isArray(val)) {
      $hiddenInput.val(val.join(','));
    } else {
      $hiddenInput.val(val || '');
    }
  });
```

Because, for single-select, val is a string like "1", not an array. Calling join on a string throws error with current patch:

Image

mmzeynalli (changes requested):

sqladmin/statics/js/main.js
mmzeynalli: Also add tests for single and multi select

mmzeynalli:
@maxim-f1 any updates?

mmzeynalli:
Im closing this, as this is embedded in #1059

PR #1044, opened 2026-05-11
Fixed overriding form_args in forms.py and widgets.py

mmzeynalli:
I stumbled upon this when I wanted to have `allow_blank` to be `True` in one of the fields. However, this was not working:

```python
form_args = {
    "personnel": {
        "allow_blank": True,
        "blank_text": "-- Choose --",
    }
}
```

Because, in sqladmin/forms.py:218:

`kwargs["allow_blank"] = nullable` overrode my param. So, I changed this (and all others) to use `setdefault`: only put values if user has not defined anything.

### DIFF
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -187,7 +187,7 @@ class ModelConverterBase:
                     else callable_default
                 )
 
-        kwargs["default"] = default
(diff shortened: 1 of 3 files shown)


aminalaee (approved):
Nice @mmzeynalli 👍

PR #1056, opened 2026-05-22
fix(list): fix the list view to limit the page width to viewport size

### LINKED ISSUE #1043: Make wide tables scrollable horizontally
### Checklist

I have tables with many columns that exceed the available viewport width. Currently, there is no visible horizontal scrollbar by default.

To scroll horizontally, I have to manually select text toward the right side of the table, which then triggers a scrollbar at the bottom of the page. This behavior makes navigating wide tables inconvenient.

A horizontal scrollbar should be visible and accessible whenever a table overflows its container width, without requiring any user interaction.

N/A

### Additional context

The table already has css class `table-responsive` which sets `overflow-x: auto`, so looks like maybe the max-width of the parent container can be adjusted to match the viewport width

CHC383:
Close: #1043

By default, flex items have `min-width: auto`, which causes the browser to calculate their minimum width based on their content. In the list view, the content inside the card container is the table itself.

Before this fix, the card containers inherited the table’s minimum width, causing overflow at the page level. As a result, the table-level `overflow-x: auto` from `.table-responsive` did not take effect.

With this fix, overflow is now correctly handled at the table level.

### DIFF
--- a/sqladmin/templates/sqladmin/list.html
+++ b/sqladmin/templates/sqladmin/list.html
@@ -3,8 +3,8 @@
 <div class="container-fluid">
   <div class="row">
     <div class="col-12">
-      <div class="d-flex">
-        <div class="flex-grow-1 me-2">
+      <div class="d-flex" style="min-width: 0;">
+        <div class="flex-grow-1 me-2" style="min-width: 0;">
           <div class="card">
             <div class="card-header">
               <h3 class="card-title">{{ model_view.name_plural }}</h3>
--- a/sqladmin/templates/sqladmin/list.html
+++ b/sqladmin/templates/sqladmin/list.html
@@ -3,8 +3,8 @@
 <div class="container-fluid">
   <div class="row">
     <div class="col-12">
-      <div class="d-flex" style="min-width: 0;">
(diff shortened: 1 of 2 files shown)


mmzeynalli (approved):

mmzeynalli:
Thanks @CHC383!

sqladmin/templates/sqladmin/list.html
```diff
@@ -3,8 +3,8 @@
 <div class="container-fluid">
   <div class="row">
     <div class="col-12">
-      <div class="d-flex">
-        <div class="flex-grow-1 me-2">
+      <div class="d-flex" style="min-width: 0;">
+        <div class="flex-grow-1 me-2" style="min-width: 0;">
```
aminalaee: Can we please move this style to css instead?

aminalaee (changes requested):

sqladmin/templates/sqladmin/list.html
CHC383: Sure, unfortunately Bootstrap and Tabler don't have a sizing utilities for these, I will create one instead

PR #1059, opened 2026-05-27
Fix/issue 1000: Fix unnecessary selectinload for AJAX relations on edit page

### LINKED ISSUE #1000: Speed slowdown issue while opening edit page in model

arturbent0:
#1000

## Problem
When opening the edit page for a model with AJAX relations (`form_ajax_refs`), sqladmin was eagerly loading all related objects via selectinload, even though they weren't needed, since the AJAX widget fetches data dynamically. This caused slow page loads when the related table had many records.

## Solution
- Skip `selectinload` for relations configured as AJAX in `form_edit_query`
- Add `get_form_data_for_edit` method that builds the form data dict manually:
  - For AJAX relations: runs a lightweight query fetching only the already-associated objects
  - For other fields: reads values directly from the model object
- Call `scaffold_form` before `get_object_for_edit` so that `_ajax_relation_names` is up to date when the query runs

## Performance
Tested with 3000 records in the related table, 2 associated to the object being edited:
- Before: 0.10s, ~219KB HTML
- After: 0.03s, ~5KB HTML

## Notes
This was tested together with #1042 (fix for "Too many fields" error). 
The diff appears large because the pre-commit hook ran `ruff format` on the files. The actual logical changes are minimal.


sqladmin/application.py
```diff
@@ -781,7 +783,7 @@ async def _handle_form_data(request: Request, obj: Any = None) -> FormData:
         there's no way to show current file of object.
         """
 
-        form = await request.form()
+        form = await request.form(max_fields=float("inf"))
```
mmzeynalli: inf can be dangerous and misused> Let's keep default value (or something bigger than that) AND have it configurable as Admin param.

mmzeynalli: And lets return 400 error when limit is exceeded.

sqladmin/models.py
```diff
@@ -1,1408 +1,1495 @@
...
+                try:
+                    count = await self._run_count(count_stmt)
+                    if count >= self.ajax_threshold:
+                        form_ajax_refs[rel.key] = create_ajax_loader(
+                            model_admin=self,
+                            name=rel.key,
+                            options={"fields": str_fields},
+                        )
+                except Exception:  # nosec B110
+                    pass
```
mmzeynalli: Do not like this. Exceptions should be caught. What can go wrong here?

sqladmin/models.py
```diff
@@ -1,1408 +1,1495 @@
...
+        pk_value = get_object_identifier(obj)
+        parent_stmt = self._stmt_by_identifier(str(pk_value))
+        for name in self._form_prop_names:
+            if name in self._ajax_relation_names:
+                relation = self._mapper.relationships[name]
+                target = relation.mapper.class_
+                direction = get_direction(relation)
+                if direction in ("ONETOMANY", "MANYTOMANY"):
+                    whereclause = parent_stmt.whereclause
+                    whereclause = parent_stmt.whereclause
```
mmzeynalli: duplicate

mmzeynalli: We can benefit from distinct on target primary key to avoid duplicate rows in edge cases.

sqladmin/models.py
```diff
@@ -1,1408 +1,1495 @@
...
+        if self.ajax_threshold > 0:
+            mapper = inspect(self.model)
+            for rel in mapper.relationships:
+                if rel.key in form_ajax_refs:
+                    continue
+                related_model = rel.mapper.class_
+                str_fields = get_str_columns(related_model)
+                if not str_fields:
+                    continue
+                count_stmt = select(func.count()).select_from(related_model)
```
mmzeynalli: Not sure I like this being run on every page load for all relationships. Why not to leave user to handle AJAX part? If it gets too much data, let them switch to ajax, it is not much of a work for us to complicate things

mmzeynalli (changes requested):

arturbent0:
Resolved the requested changes. I don't think the test errors are related to this pr, though? They didn't reproduce locally

sqladmin/application.py
```diff
@@ -680,18 +689,24 @@ async def edit(self, request: Request) -> Response:
...
 
         if request.method == "GET":
             return await self.templates.TemplateResponse(
                 request, model_view.edit_template, context
             )
 
-        form_data = await self._handle_form_data(request, model)
+        try:
+            form_data = await self._handle_form_data(
+                request, model, max_fields=self.form_max_fields
```
mmzeynalli: instead of passing param on every call (consider future additions as well) handle this in that function

sqladmin/application.py
```diff
@@ -824,14 +839,18 @@ def get_save_redirect_url(
...
+        max_fields: int = 5000,
+    ) -> FormData:
         """
         Handle form data and modify in case of UploadFile.
         This is needed since in edit page
         there's no way to show current file of object.
         """
 
-        form = await request.form()
+        form = await request.form(max_fields=max_fields)
```
mmzeynalli:
```
max_fields: int | None = None

...

max_fields = max_fields or self.form_max_fields
```

sqladmin/models.py
```diff
@@ -1,1423 +1,1473 @@
...
+                    'order_by': ('id',),
+                }
+            }
+        ```
+    """
+
+    ajax_threshold: ClassVar[int] = 500
+    """Auto-switch to AJAX select when related table exceeds this row count.
+    Set to 0 to disable automatic AJAX switching.
+    """
```
mmzeynalli: Remove this

sqladmin/models.py
```diff
@@ -1,1423 +1,1473 @@
...
+            result = session.execute(stmt)
+            return result.scalar() or 0
+
+    async def _run_count(self, stmt: ClauseElement) -> int:
+        if self.is_async:
+            async with self.session_maker(expire_on_commit=False) as session:
+                result = await session.execute(stmt)
+                return result.scalar() or 0
+        else:
+            return await anyio.to_thread.run_sync(lambda: self._run_count_sync(stmt))
```
mmzeynalli: These are leftovers as well

mmzeynalli (commented):
Merge main to your current code as well, please

arturbent0:
Sorry for the tests again. I genuinely don't know why this is happening. CI is checking 56 files while locally it's only checking 54. I've also made sure my branch is up to date with main.

mmzeynalli:
@arturbent0 I fixed the problems, but your tests are failing. Please update them accordingly.

mmzeynalli:
1. Set Git line endings:
`git config core.autocrlf input`
`git config core.eol lf`
2. Set your editor default EOL to LF.
3. Re-save changed files as LF.
4. Re-stage and recommit:
`git add --renormalize .`

Your tests are still failing

mmzeynalli:
Return uv.lock as well please

sqladmin/statics/js/main.js
```diff
@@ -122,29 +122,45 @@ $(':input[data-role="datetimepicker"]:not([readonly])').each(function () {
...
       }
     }
   });
 
-  existing_data = $(this).data("json") || [];
-  for (var i = 0; i < existing_data.length; i++) {
-    data = existing_data[i];
+  $select.on('change', function () {
+    var val = $select.val();
+    $hiddenInput.val(val ? val.join(',') : '');
```
mmzeynalli: Single-value AJAX relations can break on selection due to val.join on a string.

For non-multiple Select2 fields, `val` is a string, so val.join is undefined and can throw at runtime.
This risks breaking create/edit submissions for many-to-one AJAX fields.

I checked locally, and I got this error message in console:

Image

mmzeynalli (changes requested):

sqladmin/statics/js/main.js
arturbent0: Based on your review comment, I fixed the val.join() issue: ```Error: Uncaught Error: No Listener: tabs:outgoing.message.ready 17main.js:151 Uncaught TypeError: val.join is not a function at HTMLSelectElement.<anonymous> (main.js:151:32)
 at HTMLSelectElement.dispatch (jquery.min.js:2:43064) 
at v.handle (jquery.min.js:2:41048)
 at Object.trigger (jquery.min.js:2:71515)
 at HTMLSelectElement.<anonymous> (jquery.min.js:2:72110) 
at S.each (jquery.min.js:2:3003) 
at S.fn.init.each (jquery.min.js:2:1481) 
at S.fn.init.trigger (jquery.min.js:2:72086) 
at n.select (select2.full.min.js:2:34040)
at i.select (select2.full.min.js:2:36754)```
However, your screenshot appears to show a different error from Bootstrap. I didn't make any changes related to that Bootstrap error because I assumed it was unrelated to the val.join() issue. Please correct me if I'm mistaken and those errors are connected

mmzeynalli:
@arturbent0 I totally missed your update, my bad. Is everything ready to review? Could you also fix conflicts?

arturbent0:
> @arturbent0 I totally missed your update, my bad. Is everything ready to review? Could you also fix conflicts?

yes, its ready

mmzeynalli:
@arturbent0 your merge was not correct, and messed up `main.js`. I fixed it, also removed `hiddenInput`. I saw no value in changing input type, but if you have a valid reason, please do tell.

mmzeynalli (approved):

PR #1061, opened 2026-05-30
fix: render column defaults on the create form

### LINKED ISSUE #1060: Form field defaults are not rendered on the create page
### Checklist

### Describe the bug

Column-level field defaults (`Field(default=True)`) are silently dropped. They never reach the rendered create form for any field type. Even when surfaced via `form_args["default"]`, `BooleanField` checkboxes still render unchecked on `GET` because of how the form is constructed.

```python
import asyncio
from starlette.datastructures import FormData
from sqlmodel import SQLModel, Field
from sqladmin.forms import get_model_form

class Item(SQLModel, table=True):
    id: int = Field(primary_key=True)
    name: str = Field(default="untitled")
    priority: int = Field(default=5)
    is_active: bool = Field(default=True)

async def main():
    # No form_args - relying on the SQLModel column defaults.
    Form = await get_model_form(
        model=Item, session_maker=None, form_include_pk=False,
    )
    print("WTForms field defaults after sqladmin scaffolding:")
    for f in ("name", "priority", "is_active"):
        print(f"  {f}.default = {Form()._fields[f].default!r}")
    # name.default = None ; priority.default = None ; is_active.default = None
    # -> column-level defaults dropped for every field type.

    print("\nRendered HTML on GET (Form(empty_FormData), as application.py:create does):")
    f = Form(FormData([]))
    for n in ("name", "priority", "is_active"):
        print(f"  {n}: {getattr(f, n)()}")
    # name="", priority="", is_active unchecked

asyncio.run(main())
```

### Expected behavior

Column-level defaults (`Field(default=...)`) pre-populate the create form on `GET` for any field type. `form_args["default"]` should behave the same as a column default and work consistently across field types.

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

- Python: 3.12
- sqladmin: 0.27.0
- Framework: Starlette / FastAPI
- ORM: SQLModel on top of SQLAlchemy 2.x

### Additional context

Related:

- #753 / #756 (form_args wiring).
- #1044 same family of bug for `allow_blank`: fixed by switching direct `kwargs[...] = ...` assignments to `setdefault`

dhcsousa:
Closes #1060

Two stacked bugs prevented column-level field defaults from rendering on the create form:

1. Column defaults (e.g. `Column(default=True)`) never reached the form fields - `sqladmin` was overwriting them with `None` while building each field, so the column's own default was always discarded.
2. `application.py:create` always passed formdata to `Form()`, even on GET. WTForms treats an empty `FormData` as a real submission, so `BooleanField` / `SelectMultipleField`'s `process_formdata` then overwrites the default with `False` / `[]` before render.

Removing the early `setdefault` lets column defaults flow through `_prepare_column` for every field type. Skipping formdata on GET lets WTForms fall back to those defaults at render time.

Tests cover both layers: a unit test that checks `field.default` is populated from the column, plus end-to-end render tests on the GET create page for string, integer, and non-nullable boolean columns. There is also a test that submits the create form with the checkbox unchecked, to make sure defaults don't override real submissions.

### DIFF
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -143,7 +143,6 @@ class ModelConverterBase:
         kwargs.setdefault("label", label)
         kwargs.setdefault("validators", [])
         kwargs.setdefault("filters", [])
-        kwargs.setdefault("default", None)
         kwargs.setdefault("description", prop.doc)
         kwargs.setdefault("render_kw", widget_args)
 
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -155,6 +155,10 @@ class ModelConverterBase:
                 prop=prop, session_maker=session_maker, kwargs=kwargs, loader=loader
             )
 
+        # Fallback so kwargs always has "default"; column defaults set above win.
+        if kwargs is not None:
+            kwargs.setdefault("default", None)
+
         return kwargs
 
     def _prepare_column(
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -624,6 +624,17 @@ class Admin(BaseAdminView):
         model_view = self._find_model_view(identity)
 
         Form = await model_view.scaffold_form(model_view._form_create_rules)
+
+        if request.method == "GET":
+            form = Form()
+            context = {
+                "model_view": model_view,
+                "form": form,
+            }
+            return await self.templates.TemplateResponse(
+                request, model_view.create_template, context
+            )
+
         form_data = await self._handle_form_data(request)
         form = Form(form_data)
 
@@ -631,11 +642,6 @@ class Admin(BaseAdminView):
             "model_view": model_view,
             "form": form,
         }
-
-        if request.method == "GET":
-            return await self.templates.TemplateResponse(
-                request, model_view.create_template, context
-            )
 
         if not form.validate():
             return await self.templates.TemplateResponse(
--- a/tests/test_forms/test_forms.py
+++ b/tests/test_forms/test_forms.py
@@ -337,7 +337,7 @@ async def test_model_field_clashing_with_wtforms_reserved_attribute() -> None:
     class DataModel(Base):
         __tablename__ = "model_with_wtforms_reserved_attribute"
         id = Column(Integer, primary_key=True)
-        data = Column(String)
+        data = Column(String, default="untitled")
         errors = Column(String)
         process = Column(String)
         validate = Column(Boolean)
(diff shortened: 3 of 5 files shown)


dhcsousa:
@mmzeynalli It's done 😄, let me know what you think.

sqladmin/forms.py
```diff
@@ -143,7 +143,6 @@ async def _prepare_kwargs(
         kwargs.setdefault("label", label)
         kwargs.setdefault("validators", [])
         kwargs.setdefault("filters", [])
-        kwargs.setdefault("default", None)
```
mmzeynalli, line 146: I think we can keep this? setdefault only sets if key is missing, so if default is set, this line will do nothing. Correct me if Im wrong

mmzeynalli (commented):
Also add default to the wtform "keyword" named field, to ensure there is no error.We used to have problems with it before

sqladmin/forms.py
dhcsousa: You're right, setdefault doesn't do anything when the key already exists, but thats actually the problem - `_prepare_kwargs` sets `default` to `None` first, so later when `_prepare_column` tries to `setdefault` with the column `default`, the key is already there (as `None`), and nothing happens. Column `default` gets dropped.

You can see it in the failing `test_model_form_column_default` before the fix (first commit in this PR).

For your other point about `wtforms` needing `default` to be passed - moved the `setdefault("default", None)` to `after _prepare_column` / `_prepare_relationship` run, so `kwargs` always has `default` but column defaults can flow through.

mmzeynalli:
> Also add default to the wtform "keyword" named field, to ensure there is no error.We used to have problems with it before

What about this one?

dhcsousa:
> > Also add default to the wtform "keyword" named field, to ensure there is no error.We used to have problems with it before
> 
> What about this one?

Well, this is already handled, right? Maybe I am seeing it wrong, but the `default` fallback runs for every field before the `data -> data_` renaming happens, so keyword-named fields get a default just like any other field.

Either way, I changed an existing test (`test_model_field_clashing_with_wtforms_reserved_attribute`) to make sure this is true. gave the data column a `default="untitled"` and added an assert that it shows up on the renamed field: `assert Form()._fields["data_"].default == "untitled"`. Basically, this means that reserved name column builds fine which I think is the point.

Let me know if I am missing something, maybe I am looking at it from the wrong perspective?

mmzeynalli (approved):
Nice work!

PR #1063, opened 2026-05-31
Preserve list page after editing a record (#1055)

### LINKED ISSUE #1055: After editing, return to the last page of list view
### Discussed in https://github.com/smithyhq/sqladmin/discussions/793

<div type='discussions-op-text'>

<sup>Originally posted by **bilalovmc** July 15, 2024</sup>
After editing, return to the previous page
Now it always returns to page 1, very inconvenient
image
image
image
should have gone back to page 3

</div>

Vansh-Sharma27:
## What this does

Fixes #1055.

Right now if you are on page 3 of a list view and edit a record, after pressing **Save** you always get sent back to page 1. If you are working through a long list this is annoying because you lose your place every time.

This PR keeps the list query string (the `page`, and also sorting/search params since they live in the same query string) while you edit, and uses it for the redirect after saving, so you end up back on the page you came from.

## How

The page info was getting dropped at a few points, so I carried it through the whole flow:

- **`list.html`** – the Edit link now appends the current list query string, so opening the edit page keeps `?page=3`.
- **`edit.html`** – the form `action` and the Cancel link keep the same query string, so the POST happens on `.../edit/{pk}?page=3`.
- **`get_save_redirect_url`** – on the `Save` action it now reuses `request.url.query` when building the list URL. The other actions (continue editing / save as new / add another) are unchanged.

Only the `Save` action reads the query string; the others go to fixed edit/create pages so they ignore it.

## Tests

Extended `test_get_save_redirect_url` to check that `Save` keeps `?page=3` and that `Save and continue editing` still goes to the edit page regardless. Full test suite passes locally, and `ruff` / `ruff format` / `mypy` are clean.

### DIFF
--- a/sqladmin/templates/sqladmin/edit.html
+++ b/sqladmin/templates/sqladmin/edit.html
@@ -8,7 +8,7 @@
     </div>
     <div class="card-body border-bottom py-3">
       {% block edit_form %}
-      <form action="{{ model_view._build_url_for('admin:edit', request, obj) }}" method="POST"
+      <form action="{{ model_view._build_url_for('admin:edit', request, obj) }}{% if request.url.query %}?{{ request.url.query }}{% endif %}" method="POST"
         enctype="multipart/form-data">
         <div class="row">
           {% if error %}
@@ -20,7 +20,7 @@
         </fieldset>
         <div class="row">
           <div class="col-md-2">
-            <a href="{{ url_for('admin:list', identity=model_view.identity) }}" class="btn">
+            <a href="{{ url_for('admin:list', identity=model_view.identity) }}{% if request.url.query %}?{{ request.url.query }}{% endif %}" class="btn">
               Cancel
             </a>
           </div>
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -814,7 +814,10 @@ class Admin(BaseAdminView):
         identifier = get_object_identifier(obj)
 
         if form.get("save") == "Save":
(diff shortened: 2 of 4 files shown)


sqladmin/templates/sqladmin/edit.html
```diff
@@ -8,7 +8,7 @@ <h3 class="card-title">Edit {{ model_view.name }}</h3>
     </div>
     <div class="card-body border-bottom py-3">
       {% block edit_form %}
-      <form action="{{ model_view._build_url_for('admin:edit', request, obj) }}" method="POST"
+      <form action="{{ model_view._build_url_for('admin:edit', request, obj) }}{% if request.url.query %}?{{ request.url.query }}{% endif %}" method="POST"
```
mmzeynalli, line 11: We are passing request object here, cant we get page number from `request.referer`? If yes, it would be easier to handle

Vansh-Sharma27: Actually re-reading your point, you're right that the `Referer` on that `POST` would still have `?page=3` since the edit link already carries it in the `URL`. My reasoning there was off.
The reason I avoided `request.referer` anyway is reliability. The header can get silently stripped by `Referrer-Policy`, privacy settings, or certain browser extensions, so pagination would just break with no obvious cause. `request.url.query` is direct and the state is explicitly carried through the form action rather than depending on a header that isn't guaranteed.

mmzeynalli (approved):

PR #1064, opened 2026-06-01
Feature/litestar

BigBread57:


mmzeynalli:
Hi! Can you explain why instead of adding support to Litestar, you replaced everything starlette related with it?

PR #1071, opened 2026-06-11
build(deps): remove tabler-icons.min.css.map

CHC383:
This is outdated after tabler-icons being updated from v3.10.0 to v3.44.0, and I couldn't find a prebuilt replacement.


mmzeynalli:
What was it for before? And are you sure it is replaced in tabler-icons?

CHC383:
The CSS source map is primarily intended for debugging. When CSS source maps are enabled in the browser’s developer tools, it allows users to trace generated CSS back to the original `tabler-icons` source. However, in this case, the map only points back to a single source file, `tabler-icons.scss` (published as `dist/tabler-icons.scss` in the [package](https://www.npmjs.com/package/@tabler/icons-webfont?activeTab=code)), so I’m not sure it provides significant value.

Additionally, the project currently includes only this single source map file for `tabler-icons`, so it’s unclear whether maintaining it is worthwhile.

Since `tabler-icons` has been upgraded from v3.10.0 to v3.44.0, the existing source map is also likely out of date, meaning the mappings may no longer accurately reflect the current sources.

mmzeynalli (approved):

PR #1072, opened 2026-06-11
fix: follow symlinks for static files

### LINKED ISSUE #863: Access to statics fails when using uv in symlink mode
### Checklist

### Describe the bug

I use uv with UV_LINK_MODE=symlink (cf. <https://docs.astral.sh/uv/configuration/environment/#uv_link_mode>).
In this mode, access to statics fails because, by default, in Starlette, StaticFiles are configured with `follow_symlink: bool = False`.

Install a demo app with `UV_LINK_MODE=symlink uv sync` (or `UV_LINK_MODE=symlink uv pip install -e .` I guess).

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

Latest sqladmin, latest everything else.

### Additional context

_No response_

Dexter2099:
## Summary

Fixes static file access when SQLAdmin is installed in an environment where package files are symlinked, such as `uv` with `UV_LINK_MODE=symlink`.

SQLAdmin mounts Starlette `StaticFiles` for package statics. Starlette defaults `follow_symlink=False`, which can prevent package static assets from being served when the installed package path contains symlinks.

This change sets `follow_symlink=True` for SQLAdmin's static files mount.

## Tests

- `uv run pytest tests/test_application.py -q`
- `uv run pytest tests -q`
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run mypy sqladmin`

Fixes #863

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -418,6 +418,7 @@ class Admin(BaseAdminView):
         debug: bool = False,
         templates_dir: str = "templates",
         authentication_backend: AuthenticationBackend | None = None,
+        static_files_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """
         Args:
@@ -430,6 +431,7 @@ class Admin(BaseAdminView):
             logo_width: Width of the logo image in pixels. Defaults to 64.
             logo_height: Height of the logo image in pixels. Defaults to 64.
             favicon_url: URL of favicon to be displayed.
+            static_files_kwargs: Extra keyword arguments for Starlette StaticFiles.
         """
 
         super().__init__(
@@ -447,7 +449,9 @@ class Admin(BaseAdminView):
             authentication_backend=authentication_backend,
         )
 
-        statics = StaticFiles(packages=["sqladmin"], follow_symlink=True)
+        statics = StaticFiles(
+            **{"packages": ["sqladmin"], **(static_files_kwargs or {})}
+        )
 
         async def http_exception(
             request: Request, exc: Exception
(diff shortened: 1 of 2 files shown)


mmzeynalli (changes requested):
I believe this would break default behavior. Instead, I suggest add param (something like `static_files_kwargs`) where user can pass not only `follow_symlink`, but also other params, if they want.

Dexter2099:
I’ll preserve the default and make this opt-in via a `static_files_kwargs` parameter on `Admin`, so users who need symlink support can pass `{"follow_symlink": True}` while other `StaticFiles` kwargs remain available.

sqladmin/application.py
```diff
@@ -447,7 +449,9 @@ def __init__(  # type: ignore[no-any-unimported]
             authentication_backend=authentication_backend,
         )
 
-        statics = StaticFiles(packages=["sqladmin"])
+        statics = StaticFiles(
+            **{"packages": ["sqladmin"], **(static_files_kwargs or {})}
+        )
```
mmzeynalli: We should prevent user overriding `packages`. So it should be smth:

```python
static_files_kwargs = static_files_kwargs  or {}
static_files_kwargs['packages'] = ['sqladmin'] 
```

Something like this

mmzeynalli (changes requested):

Dexter2099:
`packages` is now applied after user-provided `static_files_kwargs`, so it cannot be overridden, and I added a regression test for that case.

mmzeynalli (approved):

PR #1074, opened 2026-06-11
feat: add rich_text_fields for pluggable rich text editors

### LINKED ISSUE #875: Add RichTextField to any field, control in ModelView
### Checklist

The current solution for CKEditor is not suitable (or reusable) if I have a model with two fields that requires CKEditor. Or any other model, that have different field name, not `content`. For each case, I need to add separate `edit_template.html` file, in order to control it for each case.

I would like to control it on the code, like

```python
class PostAdmin(ModelView, model=Post):
  rich_fields = {'content'}
```

_No response_

### Additional context

_No response_

vahidzhe:
## What this does

Adds a `rich_text_fields` attribute to `ModelView` so you can turn any
`Text` column into a WYSIWYG editor without writing custom templates.

Until now the only way to get a rich text editor was to copy the
`create.html`/`edit.html` templates into your project, extend them, and
hand-write the editor's JS. That had to be repeated for every model and
every field, and there was no way to mix different editors on the same
form. This wires the whole thing into `ModelView` directly.

Inspired by #875, but instead of hardcoding a single editor it ships a
small pluggable system so people can pick the editor they actually want.

## Usage

```python
# Simplest case - every field uses CKEditor5
class PostAdmin(ModelView, model=Post):
    rich_text_fields = ["content", "summary"]

# Or pick a different editor per field
class PostAdmin(ModelView, model=Post):
    rich_text_fields = {
        "content": CKEditor5(min_height=300),
        "summary": TinyMCE(api_key="your-key"),
    }
```

### CKEditor 5 (default)
image

### TinyMCE
image

### Quill
image

### Summernote
image

## How it works

Each editor is a `RichTextEditor` subclass that declares a Jinja2 template
and a small context dict. The create/edit pages run a fixed pipeline:
load styles, load scripts, then render each editor's init block. Because
the editor only supplies data and never touches the pipeline, adding a new
editor is just a new class plus a template.

When several fields share the same editor instance, its CDN assets are
emitted once, so you don't end up loading the same library twice (which,
for example, breaks CKEditor with a `ckeditor-duplicated-modules` error).

Only `Text`/`UnicodeText` columns are allowed - anything else raises a
`ValueError`, since the editors attach to `<textarea>` elements and other
column types don't render as textareas.

## Custom editors

Anyone can plug in their own editor by subclassing `RichTextEditor`:

```python
class MyEditor(RichTextEditor):
    template_name = "myapp/editors/my_editor.html"

    @property
    def scripts(self) -> list[str]:
        return ["https://example.com/my-editor.js"]

    def get_context(self) -> dict:
        return {"min_height": 200}
```

## Changes

- New `sqladmin/editors.py` with `RichTextEditor` base + four adapters
- `rich_text_fields` attribute and helper properties on `ModelView`
- `RichTextFieldsType` in `_types.py`
- Editor templates under `sqladmin/templates/sqladmin/editors/`
- `{% block tail %}` wired into `create.html` and `edit.html`
- Tests in `tests/test_rich_text_editors.py`
- Docs: updated `cookbook/using_wysiwyg.md`, new
  `api_reference/editors.md`, nav entry in `mkdocs.yml`

## Notes

Backward compatible - `rich_text_fields` defaults to `[]`, so existing
views are unchanged. The old "copy a template" approach still works and is
kept in the docs for advanced cases.

Closes #875


mmzeynalli:
Nice work! Although, I just realized one thing that we could do, tell me if thats possible:

Instead of different param, we use combination of  `form_overrides` and `form_args`, and use these classes there. Something like this:

```python
# Simplest case - every field uses CKEditor5
class PostAdmin(ModelView, model=Post):
    form_overrides = {'content': CKEditor5}
    form_args = {'content': 'min_height': 300}
```

I think this is better approach, as RichText is another widget of the form.

sqladmin/templates/sqladmin/create.html
```diff
@@ -41,4 +41,26 @@ <h3 class="card-title">New {{ model_view.name }}</h3>
     </div>
   </div>
 </div>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
+
+
+
+
```
mmzeynalli: too much space

sqladmin/templates/sqladmin/edit.html
```diff
@@ -45,4 +45,25 @@ <h3 class="card-title">Edit {{ model_view.name }}</h3>
     </div>
   </div>
 </div>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
+
+
+
```
mmzeynalli: same

mmzeynalli (changes requested):

vahidzhe:
> Nice work! Although, I just realized one thing that we could do, tell me if thats possible:
> 
> Instead of different param, we use combination of `form_overrides` and `form_args`, and use these classes there. Something like this:
> 
> ```python
> # Simplest case - every field uses CKEditor5
> class PostAdmin(ModelView, model=Post):
>     form_overrides = {'content': CKEditor5}
>     form_args = {'content': 'min_height': 300}
> ```
> 
> I think this is better approach, as RichText is another widget of the form.

Thanks! And you're right. I completely overlooked reusing form_overrides +
form_args. Treating rich text as just another widget that way makes much
more sense than a separate param.

One thing I want to get right before reworking it: the editors need their
CDN scripts/styles injected into the page, and shared assets loaded only
once (otherwise e.g. CKEditor throws a duplicated-modules error). WTForms
widgets normally just return the field HTML, so how would you want to
handle that part? a hook on the widget that the templates pick up, or
something else?

mmzeynalli:
Let's do smth like this:

```python
class FieldMedia:
    def __init__(self, css: list[str] = (), js: list[str] = ()):
        self.css = list(css)
        self.js = list(js)

    def __add__(self, other: "FieldMedia") -> "FieldMedia":
        seen_css = set(self.css)
        seen_js = set(self.js)
        return FieldMedia(
            css=self.css + [u for u in other.css if u not in seen_css],
            js=self.js  + [u for u in other.js  if u not in seen_js],
        )

class CKEditor5Field(TextAreaField):
    @property
    def media(self) -> FieldMedia:
        return FieldMedia(
            css=["https://cdn.ckeditor.com/ckeditor5/43.0.0/ckeditor5.css"],
            js=["https://cdn.ckeditor.com/ckeditor5/43.0.0/ckeditor5.umd.js"],
        )

...

def collect_form_media(form) -> FieldMedia:
    media = FieldMedia()
    for field in form:
        if hasattr(field, "media"):
            media = media + field.media
    return media

...

templates.env.globals["collect_form_media"] = collect_form_media
```


and in html:


```jinja
{% block head_css %}
{{ super() }}
{% set media = collect_form_media(form) if form is defined else none %}
{% if media %}
  {% for url in media.css %}<link rel="stylesheet" href="{{ url }}">{% endfor %}
{% endif %}
{% endblock %}

{% block tail_js %}
{{ super() }}
{% set media = collect_form_media(form) if form is defined else none %}
{% if media %}
  {% for url in media.js %}<script src="{{ url }}"></script>{% endfor %}
  {# per-field init templates #}
  {% for field in form if field.editor_init_template is defined %}
    {% with field = field %}{% include field.editor_init_template %}{% endwith %}
  {% endfor %}
{% endif %}
{% endblock %}
```

OR

```python
# inside create():
form = Form()
context = {
    "model_view": model_view,
    "form": form,
    "form_media": collect_form_media(form),   # ← added
}

# inside edit():
form = Form(obj=model, data=self._normalize_wtform_data(model))
context = {
    "obj": model,
    "model_view": model_view,
    "form": form,
    "form_media": collect_form_media(form),   # ← added
}
```


> [!NOTE]
> Also, instead of using CDN, download .min.css and .min.js files, and reference local files.

vahidzhe:
@mmzeynalli Nice, I really like this. The FieldMedia approach is clean and the dedup
just falls out of `__add__` for free. And yeah, making them proper WTForms
fields used through `form_overrides` + ` form_args` makes way more sense than a
separate param. I should've thought of that from the start.

Couple of things before I redo it:

For the asset injection I'd probably go with the second option (passing
form_media into the create/edit context), feels a bit cleaner than the
Jinja global and easier to test. Good with you?

On the local files vs CDN note. I'd actually lean towards keeping CDN as
the default. Vendoring the minified assets for all four (CKEditor alone is
pretty chunky) would bloat the package quite a bit, and with CDN people can
just pin whatever version they want, like `CKEditor5Field(version="43.0.0")`.
We can still let them point to local files for offline setups. Wdyt?

mmzeynalli:
Okay, lets go with CDN for now then

vahidzhe:
---

## Update: reworked to use form_overrides + form_args

Per @mmzeynalli's review, I reworked this to drop the separate
`rich_text_fields` param and instead use the existing `form_overrides` +
`form_args`, since a rich text editor is really just another widget.

New usage:

```python
from sqladmin.editors import CKEditor5Field, TinyMCEField

class PostAdmin(ModelView, model=Post):
    form_overrides = {
        "content": CKEditor5Field,
        "summary": TinyMCEField,
    }
    form_args = {
        "content": {"min_height": 300},
        "summary": {"api_key": "your-key"},
    }
```

Each editor is now a `TextAreaField` that declares its CDN assets via a
`FieldMedia` object (like Django's `Media`). `collect_form_media()` merges
them across the form with dedup, so a shared library loads only once, and
it's registered as a Jinja global so the templates compute the current
form's media directly. Kept CDN loading as we agreed, so versions can be
pinned via `form_args`.

The screenshots above still reflect how each editor looks.

vahidzhe:
@mmzeynalli  I added a cookbook page (using_wysiwyg). I also have an api_reference page
ready for the editor fields, but I noticed the existing form fields don't
have their own reference page, so I left it out for now - happy to add it
if you'd prefer. Let me know which way you want to go.

sqladmin/templates/sqladmin/editors/quill.html
```diff
@@ -0,0 +1,30 @@
...
+    textarea.dataset.quillInitialized = "1";
+
+    var container = document.createElement("div");
+    container.style.minHeight = minHeight + "px";
+    textarea.parentNode.insertBefore(container, textarea);
+    textarea.style.display = "none";
+
+    var quill = new Quill(container, { theme: theme });
+    if (textarea.value) {
+      quill.root.innerHTML = textarea.value;
```
mmzeynalli: If untrusted content can reach that field (for example user-generated HTML saved in DB), admin page load can execute attacker-controlled attributes/scripts (such as onerror) when rendering the form.

Suggested fix: sanitize before injection (for example DOMPurify) or transform stored content through a safe format (Quill Delta) and avoid raw innerHTML writes.

mmzeynalli (changes requested):

sqladmin/templates/sqladmin/editors/quill.html
vahidzhe: Good catch, thanks. Switched the load to Quill's clipboard instead of
writing innerHTML directly:

  `quill.clipboard.dangerouslyPasteHTML(textarea.value)`

mmzeynalli (approved):

mmzeynalli (approved):

PR #1075, opened 2026-06-12
fix: allow clearing nullable ajax relation fields

Dexter2099:
## Summary

This fixes nullable AJAX-backed relation fields so they can be cleared in the admin form.

`data-allow-blank="1"` was already emitted server-side for nullable single AJAX relation fields, but the Select2 AJAX init path did not consume that signal. In Select2 4.0.13, clearability for single selects depends on wiring both `allowClear` and a placeholder, so the field retained the old foreign key instead of allowing it to be removed.

## What Changed

- update the Select2 AJAX init path to read `data-allow-blank`
- apply `allowClear: true` and `placeholder: ""` only for nullable single AJAX selects
- leave multi-select AJAX fields unchanged
- add regression coverage for the rendered `data-allow-blank` contract
- add a round-trip test proving clearing the field persists `None`
- add a lightweight guard for the init-path wiring in `main.js`

## Impact

Nullable foreign key fields that use AJAX Select2 can now be cleared and saved correctly. Non-nullable and multi-select AJAX relation fields keep their existing behavior.

## Validation

- `.venv\Scripts\python.exe -m pytest tests/test_ajax.py`

### DIFF
--- a/tests/test_ajax.py
+++ b/tests/test_ajax.py
@@ -6,9 +6,11 @@ from sqlalchemy import Column, ForeignKey, Integer, String, select
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 from sqlalchemy.orm import declarative_base, relationship, selectinload
 from starlette.applications import Starlette
+from wtforms import Form
 
 from sqladmin import Admin, ModelView
 from sqladmin.ajax import create_ajax_loader
+from sqladmin.fields import AjaxSelectMultipleField
 from tests.common import async_engine as engine
 
 pytestmark = pytest.mark.anyio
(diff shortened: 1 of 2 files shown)


tests/test_ajax.py
```diff
@@ -312,3 +319,14 @@ async def test_create_and_edit_forms(client: AsyncClient) -> None:
...
+
+async def test_ajax_select2_init_respects_allow_blank_flag() -> None:
+    main_js = Path("sqladmin/statics/js/main.js").read_text(encoding="utf-8")
+
+    if '$(this).data("allowBlank")' not in main_js:
+        raise AssertionError("select2 ajax init should read data-allow-blank")
+    if "allowClear: true" not in main_js and "allowClear = true" not in main_js:
+        raise AssertionError("select2 ajax init should enable allowClear")
+    if 'placeholder: ""' not in main_js and 'placeholder = ""' not in main_js:
+        raise AssertionError("select2 ajax init should set an empty placeholder")
```
mmzeynalli: This is not a way to write tests. As long as minor refactoring happens, this will fail.

Instead, trigger allowBlank and allowClear, and load page and check for correct generated page.

mmzeynalli (changes requested):

tests/test_ajax.py
Dexter2099: Removed the source-string assertion from `tests/test_ajax.py`. The remaining coverage checks the rendered `data-allow-blank` contract and the round-trip clearing behavior without coupling the test to the exact JavaScript implementation text.

mmzeynalli:
Perfect. One small detail left before I approve.

There is no test that asserts multi-select AJAX fields do not get `allowClear` when they are also nullable. The JS condition if `(allowBlank && !isMultiple)` handles this correctly in code, but it's untested

Dexter2099:
Added a focused regression test for the nullable multi-select AJAX case. It renders an `AjaxSelectMultipleField` with `allow_blank=True` and asserts that the field remains AJAX/multiple but does not emit `data-allow-blank="1"`, so it will not get the clearability path intended only for nullable single-select AJAX fields.

mmzeynalli (approved):

PR #1076, opened 2026-06-13
fix: preserve base admin identity for polymorphic rows

### LINKED ISSUE #876: Wrong detail url in polymorphic models
### Checklist

### Describe the bug

I have polymorphic models, and when I click "view" or "edit" its base url is changed, resulting in 404.

```
from sqlalchemy.orm import DeclarativeBase

class PinnedObject(DeclarativeBase):
    __tablename__ = 'core_pinnedobject'
    id: Mapped[intpk]
    pinnedable_type: Mapped[str] = mapped_column(String)
    pinnedable_id: Mapped[int]

    __mapper_args__ = {
        'polymorphic_identity': __tablename__,
        'polymorphic_on': pinnedable_type,
    }

class NewsPinned(PinnedObject):
    __mapper_args__ = {
        'polymorphic_identity': 'news_pinned',
    }


class VacancyPinned(PinnedObject):
    __mapper_args__ = {
        'polymorphic_identity': 'vacancy_pinned',
    }
```

and when I create admin class for `PinnedObject`:

```
class PinnedObjectAdmin(ModelView, model=PinnedObject):  # type: ignore[call-arg]
    def _build_url_for(self, name, request, obj):
        # Need this bcuz of polymorphism: object can be NewsPinned or VacancyPinned
        # resulting in different generated URLs
        return request.url_for(
            name,
            identity=slugify_class_name(PinnedObject.__name__),  # the original is `slugify_class_name(obj.__class__.__name__)`
        )
```

### Expected behavior

In the list view, the base url is `pinned-object/list`. When I click to one item it should beCCC

### Actual behavior

However, link changes to `news-pinned/edit/{id}` or `vacancy-pinned/edit/{id}` resulting to 404 page. I had to overwrite `_build_url_for` function, to force class name to PinnedObject. I believe, instead of `obj.__class__.__name__` it should take `model` of the AdminView.

### Debugging material

_No response_

### Environment

- OS: Windows
- Python: 3.11
- SQLAdmin:  0.20.1

### Additional context

_No response_

Dexter2099:
Fix polymorphic model row links so they keep using the registered admin view identity.
When a ModelView is registered for a base model, SQLAdmin currently builds edit/details/delete URLs from obj.__class__.__name__. For polymorphic rows, that produces the subclass identity in the URL even though the admin view was registered under the base model identity. The generated link can then point at a route that does not exist and return 404.
This change keeps related-object links using the related object's identity, but makes current-view row action URLs use the registered admin view identity.
Fixes #876
What Changed
add a dedicated helper for current-view row URLs
keep ModelView._build_url_for() for related-object links
update ModelView._url_for_delete() to use the current view identity
update the list/details/edit templates to use the current-view helper for row actions
add a regression test covering a base-model admin rendering a polymorphic subclass row
Validation
.venv\Scripts\python.exe -m pytest -q

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -791,9 +791,7 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
     def _url_for_delete(self, request: Request, obj: Any) -> str:
         pk = get_object_identifier(obj)
         query_params = urlencode({"pks": pk})
-        url = request.url_for(
-            "admin:delete", identity=slugify_class_name(obj.__class__.__name__)
-        )
+        url = request.url_for("admin:delete", identity=self.identity)
         return str(url) + "?" + query_params
 
     def _url_for_details_with_prop(self, request: Request, obj: Any, prop: str) -> URL:
@@ -804,6 +802,13 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
 
     def _url_for_action(self, request: Request, action_name: str) -> str:
         return str(request.url_for(f"admin:action-{self.identity}-{action_name}"))
+
+    def _build_url_for_current_view(self, name: str, request: Request, obj: Any) -> URL:
+        return request.url_for(
+            name,
+            identity=self.identity,
+            pk=get_object_identifier(obj),
+        )
 
     def _build_url_for(self, name: str, request: Request, obj: Any) -> URL:
         return request.url_for(
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -791,6 +791,16 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
     def _url_for_delete(self, request: Request, obj: Any) -> str:
         pk = get_object_identifier(obj)
         query_params = urlencode({"pks": pk})
+        url = request.url_for(
+            "admin:delete", identity=slugify_class_name(obj.__class__.__name__)
+        )
+        return str(url) + "?" + query_params
+
+    # TODO: Merge current-view URL helpers into the legacy helpers in the next
+    # breaking-change window.
+    def _url_for_delete_current_view(self, request: Request, obj: Any) -> str:
+        pk = get_object_identifier(obj)
+        query_params = urlencode({"pks": pk})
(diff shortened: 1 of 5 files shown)


sqladmin/models.py
```diff
@@ -791,9 +791,7 @@ async def _run_query(self, stmt: ClauseElement) -> Any:
     def _url_for_delete(self, request: Request, obj: Any) -> str:
         pk = get_object_identifier(obj)
         query_params = urlencode({"pks": pk})
-        url = request.url_for(
-            "admin:delete", identity=slugify_class_name(obj.__class__.__name__)
-        )
+        url = request.url_for("admin:delete", identity=self.identity)
```
mmzeynalli, line 804: I like this, but this is a breaking change. This should stay as it is.

Instead create new function like build_url. Add TODO, to merge them later

tests/test_application.py
```diff
@@ -220,6 +242,43 @@ class UserAdmin(ModelView, model=User): ...
...
+
+    assert str(view._build_url_for_current_view("admin:edit", request, obj)) == (
+        "http://testserver/admin/pinned-object/edit/1"
+    )
+    assert view._url_for_delete(request, obj) == (
+        "http://testserver/admin/pinned-object/delete?pks=1"
+    )
+    assert str(view._build_url_for("admin:details", request, RelatedModel(id=2))) == (
+        "http://testserver/admin/related-model/details/2"
+    )
```
mmzeynalli: Just load pages with client, and check HTML result. This is too complicated.

mmzeynalli (changes requested):

Dexter2099:
Updated this to keep `_url_for_delete()` backwards-compatible with the existing object-class identity behavior, and added `_url_for_delete_current_view()` for the polymorphic/base-view case.

The list and details templates now use the current-view delete helper, while the legacy helper remains unchanged for external callers. I also replaced the direct URL construction test with a `TestClient` page-level test that checks the rendered HTML, plus a small compatibility check for the legacy helper behavior.

mmzeynalli (approved):

PR #1077, opened 2026-06-14
feat: pass request to column formatters

Dexter2099:
Summary
Add request-aware column formatters for list and detail views.
Column formatters can now optionally accept the current Starlette Request as a third argument. This lets formatters build URLs with request.url_for(...) while keeping existing two-argument formatters working unchanged.
Fixes #1015
What Changed
pass request into list and detail formatter rendering paths
support both (model, attribute) and (model, attribute, request) formatter signatures
keep existing formatter behavior for old call sites and two-argument formatters
document the optional request-aware formatter signature
add unit and rendered-page regression coverage for request-aware list/detail formatters
Validation
.venv\Scripts\python.exe -m pytest -q
.venv\Scripts\python.exe -m ruff check sqladmin tests docs
.venv\Scripts\python.exe -m ruff format --check sqladmin tests docs
.venv\Scripts\python.exe -m mypy sqladmin

### DIFF
--- a/sqladmin/models.py
+++ b/sqladmin/models.py
@@ -733,6 +733,12 @@ class ModelView(BaseView, metaclass=ModelViewMeta):
         self._detail_formatters = self._build_column_pairs(
             self.column_formatters_detail
         )
+        self._list_formatter_accepts_request = self._build_formatter_request_support(
+            self._list_formatters
+        )
+        self._detail_formatter_accepts_request = self._build_formatter_request_support(
+            self._detail_formatters
+        )
 
         self._form_prop_names = self.get_form_columns()
         self._form_relation_names = [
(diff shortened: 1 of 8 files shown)


mmzeynalli:
This is nice work! I will review it in detail during this week.

Would it be possible to have request as the first param, but to be backwards compatible? Logically, request is the first object, so by design it would be better. Also, is it possible to do it something like FastAPI/Flask? Having `request` name in params is enough, no matter the position (correct me if im wrong).

sqladmin/models.py
```diff
@@ -833,6 +834,36 @@ def _default_formatter(self, value: Any) -> Any:
...
+        return positional_count >= 3
+
+    def _call_formatter(
+        self,
+        formatter: Callable[..., Any],
+        obj: Any,
+        prop: str,
+        request: Request | None = None,
+    ) -> Any:
+        if request is not None and self._formatter_accepts_request(formatter):
```
mmzeynalli: For each row you check if formatter accepts request, causing overhead. I think this should be done once and stored in view level.

mmzeynalli (changes requested):
Export CSV/JSON should be considered for this feature as well.

Dexter2099:
Updated this to cache formatter request support on the view instance instead of inspecting formatter signatures for each row. Also threaded the current request through pretty CSV export so request-aware list formatters work there too.

JSON export currently uses raw `get_prop_value()` values rather than `column_formatters`, so I preserved that existing JSON behavior. If you want JSON export to apply column formatters as well, I can handle that as a separate behavior change.

mmzeynalli (approved):

PR #1083, opened 2026-07-10
[Feature] Add auto-resizing textarea widget with character counter

maxim-f1:
## Description

This PR introduces a new auto-resizing textarea widget with a dynamic character counter and maximum character validation.

### Key Features
* **Dynamic Auto-Resizing**: The textarea automatically adjusts its height dynamically when new lines are added or removed during input events.
* **Character Counter**: Displays the current character count alongside the `maxLength` limit (e.g., `140 / 400`).
* **Visual Validation & Input Blocking**: 
  * The counter highlights in redimmediately when the user reaches the maximum allowed character threshold.
  * Further text entry is blocked once the limit is hit to prevent invalid data submission.
* Compatible with SQLAlchemy's Text, BinaryLarge, and Binary types


mmzeynalli:
This is a nice feature!

However, can we:
1. Update with main?
2. Have auto-resize as a parameter? Maybe I have really long text. I would like a char count, but with auto-resize, it would take the whole page?
3. Can we also add a word count? I think all of them should be passed as params, with char count being true by default.

maxim-f1:
> 2\. Have auto-resize as a parameter? Maybe I have really long text. I would like a char count, but with auto-resize, it would take the whole page?

You can still shrink the input so it doesn't take up too much space. But if you start typing, it will expand again to fit everything.

image

maxim-f1:
> 2\. Have auto-resize as a parameter?

Yes, it will need some improvement.

maxim-f1:
> 3\. Can we also add a word count? I think all of them should be passed as params, with char count being true by default.

Theoretically, it could be added, but I'm not sure for what use cases. Databases usually have a character limit out of the box. The number of words is a fairly narrow case, so I'm not sure it'll be useful.

maxim-f1:
This can be done via form_args

```python
  form_args = {
      'text_without_limit': {
          'show_chars_count': False
      }
  }
```
image

mmzeynalli:
Okay, lets skip word count, you are right.

But I still insists auto_resize should be also a param like show_chars_count. Make default true, but, still user should be able to cusomize it.

sqladmin/widgets.py
```diff
@@ -124,3 +124,39 @@ def __call__(self, field: Field, **kwargs: Any) -> Markup:
...
+        if getattr(field, "show_chars_count", True) is True:
+            chars_count = '<div class="chars-count-label pt-1"></div>'
+        else:
+            chars_count = ""
+
+        return Markup(
+            "<textarea %s>%s</textarea>"
+            % (html_params(name=field.name, **kwargs), escape(kwargs.get("value", "")))
+            + chars_count
+        )  # nosec: markupsafe_markup_xss
```
mmzeynalli, line 211: Textareas must not have a value attribute.

Use `value = kwargs.pop("value", "")` before `html_params()`, then use `escape(value)` in the body.

sqladmin/widgets.py
```diff
@@ -124,3 +124,39 @@ def __call__(self, field: Field, **kwargs: Any) -> Markup:
...
+        for k in dir(flags):
+            if k in self.validation_attrs and k not in kwargs:
+                kwargs[k] = getattr(flags, k)
+
+        class_ = kwargs.get("class")
+        if getattr(field, "enable_autoresize", True) is True:
+            class_ = " ".join(filter(None, (class_, "autoresize-textarea")))
+            kwargs.setdefault("rows", "1")
+
+        kwargs["class"] = class_
```
mmzeynalli: This happens even when class_ is None. Skip instead.

sqladmin/widgets.py
```diff
@@ -124,3 +124,39 @@ def __call__(self, field: Field, **kwargs: Any) -> Markup:
             + str(Markup.escape(super().__call__(field, **kwargs)))
             + "</div>"
         )
+
+
+class TextAreaWidget(widgets.Input):
```
mmzeynalli: why not `widgets.TextArea`?

mmzeynalli: I think in this case you can downsize __call__ override

sqladmin/statics/js/main.js
```diff
@@ -206,3 +206,46 @@ $(':input[data-role="select2-tags"]').each(function () {
...
+
+  $textarea.on('input propertychange', function() {
+    updateTextareaHeight();
+  });
+
+  updateTextareaHeight(); // Resize on start
+});
+
+
+// Shows the number of characters in a textarea..
```
mmzeynalli: typo

mmzeynalli (commented):
Also, is there a chance to have it "default" field for Text type?

`Mapped[str] = mapped_column(String)` -> StringField
`Mapped[str] = mapped_column(Text)` -> TextField

mmzeynalli (changes requested):

mmzeynalli:
@maxim-f1 Hey, I plan to release new version tomorrow or day after, could you finalize this as well?

maxim-f1:
> Also, is there a chance to have it "default" field for Text type?

In this case, all text fields will become textareas instead of input fields. It seems this behavior is not expected for short strings of the String CHAR type.

mmzeynalli:
Yes, I suspected that, and was wondering if we can "separate" input from textarea

maxim-f1:
In my opinion, it is better to make a separate widget for String and CHAR in another MR

mmzeynalli (approved):

PR #1084, opened 2026-07-10
fix: resolve boolean input widget crash on WTForms 3.2

### LINKED ISSUE #1081: BooleanInputWidget crashes with WTForms >= 3.2 (AttributeError: 'BooleanInputWidget' object has no attribute 'validation_attrs')
### Checklist

### Describe the bug

Opening the **create** or **edit** form of any `ModelView` whose model has a
`Boolean` column returns a 500. Rendering the boolean field raises:

```
AttributeError: 'BooleanInputWidget' object has no attribute 'validation_attrs'
```

This happens with any WTForms `>= 3.2.0`, which is inside sqladmin's own
dependency range (`wtforms>=3.1,<3.3`) and therefore pulled in by default.

### Steps to reproduce

```python
from sqlalchemy import Boolean, Column, Integer

Base = declarative_base()

class Item(Base):
    __tablename__ = "item"
    enabled = Column(Boolean, default=True)   # <-- any Boolean column

class ItemAdmin(ModelView, model=Item):
    ...

# register ItemAdmin, then GET /admin/item/create (or /admin/item/edit/{id})
# -> 500
```

### Traceback (trimmed)

```
File ".../sqladmin/templates/sqladmin/_macros.html", line ..., in template
    {{ field(**extra) }}
File ".../wtforms/fields/core.py", line 164, in __call__
    return self.meta.render_field(self, kwargs)
File ".../wtforms/meta.py", line 64, in render_field
    return field.widget(field, **render_kw)
File ".../sqladmin/widgets.py", line 124, in __call__
    + str(Markup.escape(super().__call__(field, **kwargs)))
File ".../wtforms/widgets/core.py", line 176, in __call__
    if k in self.validation_attrs and k not in kwargs:
AttributeError: 'BooleanInputWidget' object has no attribute 'validation_attrs'
```

### Root cause

`BooleanInputWidget` subclasses WTForms' **base** `Input` widget:

```python
# sqladmin/widgets.py
class BooleanInputWidget(widgets.Input):
    input_type = "checkbox"
    def __call__(self, field, **kwargs):
        ...
        return Markup(... + str(Markup.escape(super().__call__(field, **kwargs))) + ...)
```

`Input.__call__` iterates the field flags with:

```python
if k in self.validation_attrs and k not in kwargs:
```

In **WTForms 3.2.0** the class-level `validation_attrs` attribute was **removed
from the base `Input` class**; it is now only defined on the concrete subclasses
(`TextInput`, `CheckboxInput`, `Select`, `FileInput`, …). Since
`BooleanInputWidget` derives from the base `Input` and does not define
`validation_attrs` itself, `super().__call__()` raises `AttributeError`.

Verified across versions (base `Input` class body):

| WTForms | base `Input` defines `validation_attrs` | result |
| --- | --- | --- |
| 3.1.x   | yes | works |
| 3.2.0 / 3.2.1 / 3.2.2 | no  | crashes |

(The other sqladmin widgets are unaffected because they subclass concrete
widgets: `DatePickerWidget`/`DateTimePickerWidget` -> `TextInput`,
`AjaxSelect2Widget`/`Select2TagsWidget` -> `Select`, `FileInputWidget` ->
`FileInput` — all of which define `validation_attrs`.)

### Suggested fix

Have `BooleanInputWidget` subclass `wtforms.widgets.CheckboxInput` instead of
the base `Input`. `CheckboxInput` already sets `input_type = "checkbox"` and
`validation_attrs = ["required", "disabled"]`:

```python
class BooleanInputWidget(widgets.CheckboxInput):
    def __call__(self, field, **kwargs):
        ...
```

(or, minimally, add `validation_attrs = ["required", "disabled"]` to the
existing class.)

### Workaround (until fixed)

```python
from sqladmin.widgets import BooleanInputWidget

if not hasattr(BooleanInputWidget, "validation_attrs"):
    BooleanInputWidget.validation_attrs = ["required", "disabled"]
```

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

- Windows
- Python 3.12
- sqladmin 0.28.0 
- WTForms 3.2.2 (also 3.2.0/3.2.1)
- FastAPI 0.139.0

### Additional context

_No response_

Vansh-Sharma27:
﻿## Problem

Opening the create/edit form of any `ModelView` whose model has a non-nullable `Boolean` column returns HTTP 500 when WTForms >= 3.2 is installed. WTForms 3.2 removed the class-level `validation_attrs` attribute from the base `Input` widget and moved it onto concrete subclasses like `CheckboxInput`. SQLAdmin's `BooleanInputWidget` subclasses the base `Input` and does not define `validation_attrs`, so `super().__call__()` raises `AttributeError: 'BooleanInputWidget' object has no attribute 'validation_attrs'` during template rendering.

The declared dependency range `wtforms>=3.1,<3.3` allows 3.2.x, so a normal `pip install sqladmin` on a fresh environment pulls the broken version.

## Root Cause

`BooleanInputWidget` in `sqladmin/widgets.py` inherits from `widgets.Input` (the abstract base) while semantically rendering a checkbox. In WTForms 3.1.x, the base `Input` class carried `validation_attrs = ["required", "disabled"]`, so attribute lookup succeeded through inheritance. WTForms 3.2 moved that attribute onto concrete widget subclasses only, leaving the base `Input` without it. Since `BooleanInputWidget` never defined its own `validation_attrs`, the lookup fails under 3.2.x.

This only affects non-nullable Boolean columns. Nullable Boolean columns go through a `SelectField` path in the converter and never reach this widget.

## Fix

Changed `BooleanInputWidget` to subclass `widgets.CheckboxInput` instead of `widgets.Input`. `CheckboxInput` is the semantically correct concrete widget and carries `validation_attrs` in both WTForms 3.1.x and 3.2.x. The existing `checked` pre-check (`if field.data: kwargs.setdefault("checked", True)`) and the SQLAdmin switch wrapper markup are preserved, so rendered output is unchanged.

The redundant `input_type = "checkbox"` class attribute was removed since `CheckboxInput` already defines it.

## Why not the other options

- **Adding `validation_attrs` directly to `BooleanInputWidget`** would fix the crash but keeps the semantically incorrect dependency on the base `Input` class and copies upstream checkbox policy, which could diverge on future WTForms changes.
- **Pinning `wtforms<3.2`** would reject a version currently declared as supported and block users from receiving WTForms fixes.

Subclassing the concrete `CheckboxInput` is the smallest complete fix that follows the upstream API as intended.

## Changes

- `sqladmin/widgets.py`: `BooleanInputWidget` now subclasses `widgets.CheckboxInput`; removed redundant `input_type = "checkbox"`.
- `tests/test_fields.py`: added `test_boolean_field` covering unchecked render, checked render, and `required` flag propagation through `validation_attrs` (the exact path that crashed on 3.2.x).
- `uv.lock`: bumped WTForms from 3.1.2 to 3.2.2 so CI exercises the supported upper bound.
- `.github/workflows/test-suite.yml`: added a `wtforms-compat` CI job with a two-entry matrix (WTForms 3.1 and 3.2) that runs the focused WTForms-relevant test files (`test_fields.py`, `test_forms/test_forms.py`, `test_view_sync.py`) against SQLite. Each entry uses `uv run --no-sync` after the version override and prints the installed WTForms version so the CI log proves which version actually ran.

## Verification

- Reproduced the crash on WTForms 3.2.2 before the fix (3 existing tests + the new test all failed with the reported `AttributeError`).
- After the fix: full test suite passes on WTForms 3.2.2 (378 passed, 4 skipped) with 98% coverage.
- Backward compatibility confirmed on WTForms 3.1.2 (new test + existing checkbox endpoint tests pass).
- The three focused WTForms test files pass on both WTForms 3.1.2 and 3.2.2.
- `ruff check`, `ruff format --check`, `mypy`, and `bandit` all pass clean.

Closes #1081

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -70,71 +70,15 @@ jobs:
         with:
           files: coverage.xml
 
-  discover-wtforms:
-    name: "Discover WTForms versions"
-    runs-on: "ubuntu-latest"
-    outputs:
-      matrix: ${{ steps.discover.outputs.matrix }}
-
-    steps:
-      - uses: "actions/checkout@v3"
-      - uses: "actions/setup-python@v4"
-        with:
-          python-version: "3.12"
-      - name: "Discover supported WTForms versions"
-        id: discover
-        run: |
-          python << 'PYEOF'
-          import re, json, os, tomllib
-
-          with open("pyproject.toml", "rb") as f:
-              dependencies = tomllib.load(f)["project"]["dependencies"]
-
-          wtforms_spec = None
-          for dep in dependencies:
-              if dep.split()[0].lower() == "wtforms":
-                  wtforms_spec = dep
-                  break
-          if wtforms_spec is None:
-              raise SystemExit("wtforms dependency not found in pyproject.toml")
-
-          lower = re.search(r">=\s*(\d+)\.(\d+)", wtforms_spec)
-          upper = re.search(r"<\s*(\d+)\.(\d+)", wtforms_spec)
-          if not lower or not upper:
-              raise SystemExit(f"could not parse bounds from spec: {wtforms_spec}")
-
-          lower_major, lower_minor = int(lower.group(1)), int(lower.group(2))
-          upper_major, upper_minor = int(upper.group(1)), int(upper.group(2))
-
-          if lower_major != upper_major:
-              raise SystemExit(
-                  "WTForms discovery expects lower and upper bounds "
-                  "within the same major version"
-              )
-
-          entries = []
-          minor = lower_minor
-          while minor < upper_minor:
-              entries.append({
-                  "wtforms-version": f"{lower_major}.{minor}",
-                  "wtforms-spec": f"wtforms>={lower_major}.{minor},<{lower_major}.{minor + 1}",
-              })
-              minor += 1
-
-          if not entries:
-              raise SystemExit("no WTForms versions discovered; check dependency bounds")
-
-          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
-              f.write(f"matrix={json.dumps({'include': entries})}\n")
-          PYEOF
-
   wtforms-compat:
-    name: "WTForms ${{ matrix.wtforms-version }}"
+    name: "WTForms ${{ matrix.wtforms-spec }}"
     runs-on: "ubuntu-latest"
-    needs: discover-wtforms
 
     strategy:
-      matrix: ${{ fromJSON(needs.discover-wtforms.outputs.matrix) }}
+      matrix:
+        wtforms-spec:
+          - "wtforms>=3.1,<3.2"
+          - "wtforms>=3.2,<3.3"
 
     steps:
       - uses: "actions/checkout@v3"
@@ -147,10 +91,8 @@ jobs:
           python-version: "3.12"
       - name: "Install dependencies"
         run: uv sync --all-groups
-      - name: "Install WTForms ${{ matrix.wtforms-version }}"
+      - name: "Install WTForms"
         run: uv pip install "${{ matrix.wtforms-spec }}"
-      - name: "Show WTForms version"
-        run: uv run --no-sync python -c "import wtforms; print(wtforms.__version__)"
       - name: "Run WTForms compatibility tests"
         env:
           TEST_DATABASE_URI_SYNC: "sqlite:///test.db?check_same_thread=False"
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -69,17 +69,54 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           files: coverage.xml
+
+  discover-wtforms:
+    name: "Discover WTForms versions"
+    runs-on: "ubuntu-latest"
+    outputs:
+      matrix: ${{ steps.discover.outputs.matrix }}
+
+    steps:
+      - uses: "actions/checkout@v3"
+      - uses: "actions/setup-python@v4"
+        with:
+          python-version: "3.12"
+      - name: "Discover supported WTForms versions"
+        id: discover
+        run: |
+          python << 'PYEOF'
+          import re, json, os
+          content = open("pyproject.toml").read()
+          match = re.search(r'"wtforms\s*([^"]+)"', content)
+          if not match:
+              raise SystemExit("wtforms dependency not found in pyproject.toml")
+          spec = match.group(1).strip()
+          lower = re.search(r">=\s*(\d+)\.(\d+)", spec)
+          upper = re.search(r"<\s*(\d+)\.(\d+)", spec)
+          if not lower or not upper:
+              raise SystemExit(f"could not parse bounds from spec: {spec}")
+          lower_major, lower_minor = int(lower.group(1)), int(lower.group(2))
+          upper_major, upper_minor = int(upper.group(1)), int(upper.group(2))
+          entries = []
+          major, minor = lower_major, lower_minor
+          while (major, minor) < (upper_major, upper_minor):
+              entries.append({
+                  "wtforms-version": f"{major}.{minor}",
+                  "wtforms-spec": f"wtforms>={major}.{minor},<{major}.{minor + 1}",
+              })
+              minor += 1
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
+              f.write(f"matrix={json.dumps(entries)}\n")
+          PYEOF
+
   wtforms-compat:
     name: "WTForms ${{ matrix.wtforms-version }}"
     runs-on: "ubuntu-latest"
+    needs: discover-wtforms
 
     strategy:
       matrix:
-        include:
-          - wtforms-version: "3.1"
-            wtforms-spec: "wtforms>=3.1,<3.2"
-          - wtforms-version: "3.2"
-            wtforms-spec: "wtforms>=3.2,<3.3"
+        include: ${{ fromJSON(needs.discover-wtforms.outputs.matrix) }}
 
     steps:
       - uses: "actions/checkout@v3"
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -70,8 +70,16 @@ jobs:
         with:
           files: coverage.xml
   wtforms-compat:
-    name: "WTForms upper-bound (>=3.2,<3.3)"
+    name: "WTForms ${{ matrix.wtforms-version }}"
     runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        include:
+          - wtforms-version: "3.1"
+            wtforms-spec: "wtforms>=3.1,<3.2"
+          - wtforms-version: "3.2"
+            wtforms-spec: "wtforms>=3.2,<3.3"
 
     steps:
       - uses: "actions/checkout@v3"
@@ -84,10 +92,16 @@ jobs:
           python-version: "3.12"
       - name: "Install dependencies"
         run: uv sync --all-groups
-      - name: Force WTForms upper-bound
-        run: uv pip install "wtforms>=3.2,<3.3"
-      - name: "Run tests with SQLite"
+      - name: "Install WTForms ${{ matrix.wtforms-version }}"
+        run: uv pip install "${{ matrix.wtforms-spec }}"
+      - name: "Show WTForms version"
+        run: uv run --no-sync python -c "import wtforms; print(wtforms.__version__)"
+      - name: "Run WTForms compatibility tests"
         env:
           TEST_DATABASE_URI_SYNC: "sqlite:///test.db?check_same_thread=False"
           TEST_DATABASE_URI_ASYNC: "sqlite+aiosqlite:///test.db?check_same_thread=False"
-        run: make test
+        run: |
+          uv run --no-sync pytest \
+            tests/test_fields.py \
+            tests/test_forms/test_forms.py \
+            tests/test_views/test_view_sync.py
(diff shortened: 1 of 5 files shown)


.github/workflows/test-suite.yml
```diff
@@ -69,3 +69,25 @@ jobs:
...
+          python-version: "3.12"
+      - name: "Install dependencies"
+        run: uv sync --all-groups
+      - name: Force WTForms upper-bound
+        run: uv pip install "wtforms>=3.2,<3.3"
+      - name: "Run tests with SQLite"
+        env:
+          TEST_DATABASE_URI_SYNC: "sqlite:///test.db?check_same_thread=False"
+          TEST_DATABASE_URI_ASYNC: "sqlite+aiosqlite:///test.db?check_same_thread=False"
+        run: make test
```
mmzeynalli: Can we do the following?

For the wtforms we support, we run 2-3 test files which tests wtforms (test_fields.py for example), and have that as part of CI

mmzeynalli (changes requested):

.github/workflows/test-suite.yml
```diff
@@ -69,3 +69,39 @@ jobs:
...
+    name: "WTForms ${{ matrix.wtforms-version }}"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        include:
+          - wtforms-version: "3.1"
+            wtforms-spec: "wtforms>=3.1,<3.2"
+          - wtforms-version: "3.2"
+            wtforms-spec: "wtforms>=3.2,<3.3"
```
mmzeynalli: Any way we can do this more automatic? Like when we upgrade wtforms again to 3.3, this works without change

Vansh-Sharma27: GitHub Actions evaluates matrix values when the workflow is parsed, so the current job cannot read the WTForms range from `pyproject.toml` within its own `matrix.include` block.

One possible way to make this automatic is a two-job setup:

1. A lightweight discovery job reads the WTForms constraint from `pyproject.toml` (currently `>=3.1,<3.3`), enumerates the supported minor versions (`3.1, 3.2`), and exposes the resulting matrix as JSON through `$GITHUB_OUTPUT`.
2. The `wtforms-compat` job consumes that output using `strategy.matrix.wtforms-version: ${{ fromJSON(needs.discover.outputs.versions) }}` and runs the focused test files once for each generated version range.

For example, the current constraint would generate jobs for WTForms 3.1 and 3.2. If it is later changed to `>=3.1,<3.4`, the workflow would automatically add 3.3 without requiring another manual matrix update.

The main benefit is that `pyproject.toml` remains the single source of truth and future WTForms range changes automatically update the CI coverage. The trade-off is a slightly more complex workflow because it adds one small discovery job and a short Python script for parsing the constraint and generating the matrix.

If this approach matches what you had in mind, I’d be happy to implement it.

mmzeynalli: Let's do that

Vansh-Sharma27: It's done!

JartanFTW:
Embedding extensive python code into test-suite.yml is unprecedented in sqladmin as of yet. Not to mention the creation of 2 new jobs in what is currently a relatively thin and clearly-purposed workflow. I don't see why wtforms should be given this first-class placement in the workflow diverging from all other project dependencies. It would be more appropriate to find a pytest solution in alignment with the existing testing model for the project. Wouldn't simple canary tests for wtforms behaviour be sufficient for coverage and optimal for repo consistency?

Vansh-Sharma27:
Thanks for the feedback @JartanFTW

I understand the concern about keeping the workflow consistent and avoiding dependency-specific CI complexity. Since this is a broader testing-strategy decision, I’ll defer to @mmzeynalli on whether the PR should retain the multi-version compatibility jobs or instead rely on focused canary tests within the existing test suite. I’m happy to adjust the implementation accordingly.

mmzeynalli:
To be honest, I also did not like this result, was expecting smth lighter. Lets go back:

```yaml
wtforms-compat:
  strategy:
    matrix:
      wtforms-spec:
        - "wtforms>=3.1,<3.2"
        - "wtforms>=3.2,<3.3"
  steps:
    - run: uv sync --all-groups
    - run: uv pip install "${{ matrix.wtforms-spec }}"
    - run: uv run pytest ...
```

mmzeynalli (changes requested):

Vansh-Sharma27:
﻿Makes sense. I've removed the discovery job and the embedded Python entirely and gone back to the simple static `wtforms-spec` matrix you outlined. The job runs the three focused WTForms test files (`test_fields.py`, `test_forms/test_forms.py`, `test_view_sync.py`) against each version on SQLite, with `--no-sync` so the matrix install isn't reverted by the lockfile.

mmzeynalli (approved):

JartanFTW:
This is still an entire first-class job addition to the test workflow to examine wtforms backwards compatibility. Maybe this project just hasn't encountered a backwards compat issue yet, I'm unsure, but there's no precedent for including dependency specific jobs like this in the workflow. If this is how backwards compat should be handled in the project from hereon, then some sort of standard should be defined in writing to prevent standards drifting.

mmzeynalli:
I was considering this for more like dependabot updates. This way, we could be aware if there is a breaking change of which we are not aware, as we heavily depend on it.

JartanFTW:
@mmzeynalli Since dependabot is already configured and the project uses uv.lock, removing the <3.3 ceiling on wtforms (leaving >=3.1) would let dependabot bump the lockfile pin when WTForms 3.3 drops. The full test suite, now including test_boolean_field exercising the validation_attrs codepath that broke, would run on that PR. If it fails, the PR blocks. This should fulfill your stated goal of catching breaking upstream changes automatically without the wtforms-compat CI job running on every push/PR, unless I'm missing something.

mmzeynalli:
Yep, you are right, I completely forgot that Dependabot also runs test suite. Thanks for reminding. @Vansh-Sharma27 pls remove workflow changes completely

JartanFTW:
Note that the cap on the version also must be removed for this to work properly, I believe. Otherwise dependabot doesn't try to update it. I could be wrong, I don't usually use caps.

mmzeynalli:
Do you mean `<3.3`?

mmzeynalli:
Dependabot updates that too, when time comes

JartanFTW:
Then should be all good to just remove the workflow job.

Vansh-Sharma27:
I can remove all the workflow changes as @mmzeynalli requested, but there's one thing I want to flag about the Dependabot approach first.

The repo's `dependabot.yml` uses `package-ecosystem: "pip"`, which updates `pyproject.toml` but not `uv.lock`. I went through all 17 Dependabot PRs in the repo's history (from #852 to #1079) every one of them only modified `pyproject.toml`, none touched `uv.lock`. This is because `uv.lock` support requires `package-ecosystem: "uv"`, which is a separate ecosystem ([see GitHub's supported ecosystems list](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories)).

The problem: CI runs `uv sync`, which installs from `uv.lock`, not `pyproject.toml`. Per [uv's locking docs](https://docs.astral.sh/uv/concepts/projects/sync/#checking-the-lockfile): "if you change the version constraints such that the existing locked version is still included, the lockfile will still be considered up-to-date." So when Dependabot widens the WTForms cap (e.g., from `<3.3` to `<3.4`), the lockfile stays pinned to the older version and is still treated as valid. As a result, `uv sync` installs the old version and CI never tests the new one.

This is exactly what happened with PR #1067: Dependabot widened the cap from `<3.2` to `<3.3` on June 8, `uv.lock` stayed at 3.1.2, and the 3.2 breakage wasn't caught until issue #1081 was reported on July 8 — a full month later.

Astral's [Dependabot integration guide](https://docs.astral.sh/uv/guides/integration/dependabot/) recommends adding `package-ecosystem: "uv"` to `dependabot.yml` so Dependabot updates `uv.lock` as well. That would close this gap, Dependabot would bump both the constraint in `pyproject.toml` and the pin in `uv.lock`, and CI would actually test the new version.

Shall I add `package-ecosystem: "uv"` to `dependabot.yml` in this PR to close this gap, while simultaneously reverting all the workflow changes?

JartanFTW:
Great catch on the missing uv package ecosystem for dependabot. I support adding that. I'm unsure how involved any remaining pip support is in the project, but I'd suggest standardizing everything to uv OR pip, not both, and there seems to be some mixing here. Deferring to @mmzeynalli

mmzeynalli:
Definitely uv, I worked my butt off to change poetry to uv 😉. @Vansh-Sharma27 fix it to uv please

Vansh-Sharma27:
@mmzeynalli, do you mean I should standardize everything in the repo from pip to uv, or just add `package-ecosystem: "uv"` to `dependabot.yml`?

I couldn't clearly understand what exactly you requested 😅

mmzeynalli:
Yes, lets do uv

Vansh-Sharma27:
﻿Done. I've removed the `wtforms-compat` CI job entirely and switched `dependabot.yml` from `package-ecosystem: "pip"` to `"uv"`.

The `uv` ecosystem updates both `pyproject.toml` and `uv.lock` in the same PR (I verified this against the [dependabot-core source](https://github.com/dependabot/dependabot-core/blob/main/uv/lib/dependabot/uv/file_updater.rb) — its `FileUpdater` calls both `updated_pip_compile_based_files` and `updated_uv_lock_files`, and the `LockFileUpdater` runs `uv lock --upgrade-package <name>==<version>` to regenerate the lockfile). So future Dependabot PRs for WTForms (and all other deps) will bump both the constraint and the lockfile pin, and CI will actually test the new version.

PR #1091, opened 2026-07-15
added keywords arguments for self.templates.TemplateResponse

haykeminyan:
Update here https://fastapi.tiangolo.com/advanced/templates/ now templateResponse demands use

this structure

```
templates.TemplateResponse(
        request=request, name="item.html", context={"id": id}
    )
```

previously was

```
templates.TemplateResponse(
       request, "item.html",{"id": id}
    )
```

### DIFF
--- a/docs/cookbook/displaying_one_time_secrets.md
+++ b/docs/cookbook/displaying_one_time_secrets.md
@@ -43,7 +43,7 @@ built-in create/edit handlers stamp anti-cache headers on the response
 automatically; custom views must do so explicitly:
 
 ```python
-response = await self.templates.TemplateResponse(request, "your_template.html", context)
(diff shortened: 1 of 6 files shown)


mmzeynalli:
Nice detail! If possible, do fix lint errors as well

haykeminyan:
> Nice detail! If possible, do fix lint errors as well

Thanks! added linter fixes with .pre-commit-config.yaml

mmzeynalli (changes requested):
I think there is minor issue. `TemplateResponse` we have in the code is not Starlette/FastAPI’s, but our own, in `templating.py` file. So, I believe, you only need update it there, and no need to touch anywhere else.

PR #1092, opened 2026-07-16
fix: unwrap Enum values in get_object_identifier

### LINKED ISSUE #955: bug: get_object_identifier returns Enum instance instead of string value for Enum primary keys
### Checklist

## Summary

When using SQLAdmin with SQLModel/SQLAlchemy models that have Enum primary keys, the `get_object_identifier()` helper function returns the Enum instance instead of its string value. This causes issues with URL generation, e.g. I am seeing non-working URLs like `http://localhost:8000/admin/animal/edit/AnimalEnum.DOG`

## Expected Behavior

`get_object_identifier()` should return the string value of an enum when used as a primary key, suitable for use in URLs and identifiers.

## Actual Behavior

`get_object_identifier()` returns the enum instance itself, which when converted to string gives the enum name (e.g., `"AnimalEnum.DOG"`) rather than the enum value (e.g., `"dog"`).

## Minimal Reproduction Case

```python
from enum import Enum
from sqlmodel import Field, SQLModel, create_engine, Session
from sqlmodel import Enum as DbEnum
from sqladmin.helpers import get_object_identifier

# Define the enum
class AnimalEnum(str, Enum):
    DOG = "dog"
    CAT = "cat"

# Define the model with enum primary key
class Animal(SQLModel, table=True):
    id: AnimalEnum = Field(
        primary_key=True, 
        sa_type=DbEnum(AnimalEnum, name="animalenum")
    )
    name: str

# Create a test instance
engine = create_engine("sqlite:///test.db")
SQLModel.metadata.create_all(engine)

with Session(engine) as session:
    dog = Animal(id=AnimalEnum.DOG, name="Rover")
    session.add(dog)
    session.commit()
    session.refresh(dog)
    
    # Demonstrate the issue
    identifier = get_object_identifier(dog)
    print(f"get_object_identifier(dog): {identifier}")
    print(f"str(get_object_identifier(dog)): {str(identifier)}")
    print(f"type(get_object_identifier(dog)): {type(identifier)}")
    print(f"Expected value: {dog.id.value}")
```

**Output:**
```
get_object_identifier(platform): <AnimalEnum.DOG: 'dog'>
str(get_object_identifier(platform)): AnimalEnum.DOG
type(get_object_identifier(platform)): <enum 'AnimalEnum'>
Expected value: dog
```

## Expected Output

```
get_object_identifier(dog): dog
str(get_object_identifier(dog)): dog
type(get_object_identifier(dog)): <class 'str'>
Expected value: dog
```

## Proposed Fix

The `get_object_identifier()` function should check if the primary key value is an Enum instance and return its `.value` attribute:

```python
def get_object_identifier(obj):
    # ... existing code ...
    if isinstance(pk_value, Enum):
        return pk_value.value
    return pk_value
```

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

- MacOS, Python 3.13, sqladmin 0.21.0

### Additional context

_No response_

pctablet505:
## Summary

`get_object_identifier()` returns the raw primary key value(s) for use in URLs and ajax lookups. For models whose primary key is an `Enum` column, it was returning the Enum member itself rather than its value, so `str()` on it produced `AnimalEnum.DOG` instead of `dog`. This broke generated edit/detail URLs for such models (e.g. `/admin/animal/edit/AnimalEnum.DOG`).

This unwraps Enum members to their `.value` before use, for both single and composite primary keys.

Closes #955

## Test plan

- Added `test_single_pk_identifier_with_enum` in `tests/test_helpers.py`, covering a model with an `Enum` primary key.
- `uv run pytest --cov=sqladmin --cov-report=term-missing` — 484 passed, 4 skipped (pre-existing), coverage 97.10%.
- `make lint` (ruff check, ruff format --check, mypy) — all pass.

### DIFF
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -298,6 +298,7 @@ def get_object_identifier(obj: Any) -> Any:
     """Returns a value that uniquely identifies this object."""
     primary_keys = get_primary_keys(obj)
     values = [getattr(obj, pk.name) for pk in primary_keys]
+    values = [v.value if isinstance(v, enum.Enum) else v for v in values]
 
     # Unaltered value for tables with a single primary key
     if len(values) == 1:
--- a/sqladmin/helpers.py
+++ b/sqladmin/helpers.py
@@ -294,11 +294,24 @@ def get_primary_keys(model: Any) -> tuple[Column, ...]:
     return tuple(sa_inspect(model).mapper.primary_key)
 
 
+def _coerce_pk_value(value: Any) -> Any:
+    """Unwrap an Enum primary key to its underlying value.
+
+    SQLAlchemy returns the Enum member itself (e.g. ``Color.RED``) for a
+    column typed as an Enum, not its underlying value (e.g. ``"red"``).
+    Object identifiers are embedded in admin URLs, so the raw member's
+    ``str()`` (``"Color.RED"``) would produce a broken, non-round-trippable
+    path. Non-Enum values pass through unchanged.
+    """
+    if isinstance(value, enum.Enum):
+        return value.value
+    return value
+
+
 def get_object_identifier(obj: Any) -> Any:
     """Returns a value that uniquely identifies this object."""
     primary_keys = get_primary_keys(obj)
-    values = [getattr(obj, pk.name) for pk in primary_keys]
-    values = [v.value if isinstance(v, enum.Enum) else v for v in values]
+    values = [_coerce_pk_value(getattr(obj, pk.name)) for pk in primary_keys]
 
     # Unaltered value for tables with a single primary key
     if len(values) == 1:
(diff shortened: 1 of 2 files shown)


mmzeynalli:
Interesting. Is it somehow common to have enum field as primary key? PR seems okay, but I find it obsolete, as this is not standard and not good practice to use enum as pk

mmzeynalli (approved):
A few optional follow-ups before/after merge, none blocking:

- Could we add a case for a plain (non-str) enum, e.g. class Color(enum.Enum): RED = "red"? That's the more visibly broken scenario from the docstring — Color.RED != "red" and str(Color.RED) == "Color.RED" — so it exercises the fix more directly than the str-mixed enum.
- A composite-PK case with an enum would cover the multi-PK ";".join(...) branch, which the coercion touches but no test hits yet.
- This is a unit test of the outbound direction only (object → identifier). Since #955 is really about routes breaking, an inbound assertion — e.g. object_identifier_values("dog", Animal) resolving back to the enum — would prove the URL is genuinely round-trippable, which is what the docstring promises.

PR #1097, opened 2026-07-18
fix: preserve wall-clock time for DateTime(timezone=True) columns on edit

### LINKED ISSUE #796: Datetime with timezone edited incorrectly
### Checklist

### Describe the bug

Datetime with timezone are edited incorrectly. In this example I have a table with a name and a created_at column: 

```python
class User(Base):
    __tablename__ = "users"

    id: Mapped[int] = mapped_column(primary_key=True)
    name: Mapped[str]
    created_at: Mapped[dt.datetime] = mapped_column(
        DateTime(timezone=True), server_default=func.now()
    )
```
 
If I change the name of a row, it will also edit the created_at column, most likely because the timezone is missing from the edit field: 

Screenshot 2024-07-23 at 11 55 18

Screenshot 2024-07-23 at 11 55 30

Screenshot 2024-07-23 at 11 55 35

You can see that the `created_at` column gets updated (from 9:55 to 7:55)

Here is the complete program used in the example:

```python
from contextlib import asynccontextmanager
import datetime as dt
from sqlalchemy import DateTime, func, text
from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
from sqlalchemy.orm import Mapped, declarative_base, mapped_column

from fastapi import FastAPI


Base = declarative_base()
engine = create_async_engine(
    "postgresql+asyncpg://wave:@localhost/sqladmin-testing",
)


class User(Base):
    __tablename__ = "users"

    id: Mapped[int] = mapped_column(primary_key=True)
    name: Mapped[str]
    created_at: Mapped[dt.datetime] = mapped_column(
        DateTime(timezone=True), server_default=func.now()
    )


Session = async_sessionmaker(bind=engine)


@asynccontextmanager
async def lifespan(app: FastAPI):
    async with engine.begin() as conn:
        await conn.run_sync(Base.metadata.create_all)

    async with Session() as conn:
        await conn.execute(text("INSERT INTO users (name) values ('foo'), ('bar')"))
        await conn.commit()
    yield


app = FastAPI(lifespan=lifespan)
admin = Admin(app, engine)


    column_list = [User.id, User.name, User.created_at]


admin.add_view(UserAdmin)
```

The issue doesn't happen with SQLite, only with PostgreSQL

### Expected behavior

I expect the `created_at` column to not be modified.

### Actual behavior

The `created_at` column gets edited incorrectly (from 9:55 to 7:55)

### Debugging material

_No response_

### Environment

sqladmin version: 0.18.0

### Additional context

_No response_

AdarshJ173:
## Summary

Closes #796

Editing a row with a `DateTime(timezone=True)` column silently shifts the stored time on every save — e.g. `09:55 +02:00` becomes `07:55 UTC` — because WTForms' base `DateTimeField` renders the value without its UTC offset, and then re-parses the submitted string as naive UTC.

---

## Root Cause

`ModelConverter.conv_datetime` always returns a plain `DateTimeField`. WTForms renders `self.data` via `datetime.strftime(fmt)` which **drops `tzinfo`** entirely. The HTML `<input type="datetime-local">` therefore shows `09:55:00` regardless of whether the original value was `09:55:00+02:00`. On submit, WTForms parses `09:55:00` as a naive datetime; PostgreSQL interprets it as UTC and stores `07:55:00 UTC` — a two-hour drift equal to the original offset.

This only manifests with PostgreSQL (or any backend that returns tz-aware datetimes); SQLite returns naive datetimes so the bug is invisible there — matching the reporter's observation.

---

### `sqladmin/fields.py` — new `DateTimeLocalField`

```python
class DateTimeLocalField(DateTimeField):
    def __init__(self, *args, timezone: bool = False, **kwargs):
        super().__init__(*args, **kwargs)
        self._timezone = timezone

    def process_data(self, value):
        # Normalise tz-aware → naive UTC before display
        if value is not None and value.tzinfo is not None:
            value = value.astimezone(dt_timezone.utc).replace(tzinfo=None)
        super().process_data(value)

    def process_formdata(self, valuelist):
        super().process_formdata(valuelist)          # parses to naive datetime
        if self._timezone and self.data is not None:
            self.data = self.data.replace(tzinfo=dt_timezone.utc)  # re-attach UTC
```

### `sqladmin/forms.py` — `conv_datetime` uses `DateTimeLocalField` for tz columns

```python
@converts("DateTime")
def conv_datetime(self, model, prop, kwargs):
    column = prop.columns[0]
    if getattr(column.type, "timezone", False):
        kwargs["timezone"] = True
        return DateTimeLocalField(**kwargs)
    return DateTimeField(**kwargs)   # unchanged for naive columns
```

---

## Behaviour After Fix

| Scenario | Before | After |
|----------|--------|-------|
| Load row with `09:55:00+02:00` | Form shows `09:55:00`, saves as `07:55 UTC` ❌ | Form shows `09:55:00 UTC`, saves as `09:55 UTC` ✅ |
| Load row with naive datetime (SQLite) | `09:55:00`, saves as `09:55` ✅ | Unchanged ✅ |
| Non-tz `DateTime()` column | `DateTimeField` used | `DateTimeField` used, unchanged ✅ |

> **Note on display:** The field normalises to UTC before rendering. If your app stores times in a non-UTC timezone, the displayed value will be the UTC equivalent. This is the safest, unambiguous representation for a timezone-unaware HTML input and is consistent with how other admin frameworks (Flask-Admin, Django admin) handle this.

---

## Checklist

- [x] No breaking changes for naive `DateTime()` columns (code path unchanged)
- [x] No new dependencies
- [x] Pure Python stdlib (`datetime.timezone.utc`) — no third-party tz library required
- [x] `DateTimeLocalField` exported from `sqladmin.fields.__all__`

### DIFF
--- a/sqladmin/fields.py
+++ b/sqladmin/fields.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import operator
+from datetime import timezone as dt_timezone
 from enum import Enum
 from typing import Any, Callable, Generator
 from uuid import UUID
@@ -28,6 +29,7 @@ __all__ = [
     "CDNURLField",
     "DateField",
     "DateTimeField",
+    "DateTimeLocalField",
     "FileField",
     "IntervalField",
     "JSONField",
@@ -57,6 +59,56 @@ class DateTimeField(fields.DateTimeField):
     """
 
     widget = sqladmin_widgets.DateTimePickerWidget()  # type: ignore[assignment]
+
+
+class DateTimeLocalField(DateTimeField):
+    """
+    Variant of :class:`DateTimeField` for ``DateTime(timezone=True)`` columns.
+
+    PostgreSQL (and other timezone-aware backends) return tz-aware
+    ``datetime`` objects.  WTForms renders them by calling
+    ``datetime.strftime`` which silently drops the UTC offset, so the
+    value that ends up in the HTML input is the *wall-clock* time in the
+    database's stored timezone.  When the user submits the form without
+    touching the field the value is parsed as a *naive* datetime and
+    written back to the DB, which then interprets it as UTC – causing an
+    apparent time-shift equal to the original UTC offset.
+
+    This field fixes the round-trip by:
+
+    1. **process_data** – if the incoming value is tz-aware, convert it
+       to UTC and then make it naive before storing in ``self.data``.  The
+       HTML input therefore always shows a UTC wall-clock time, which is
+       unambiguous and stable.
+    2. **process_formdata** – after the user submits, re-attach
+       ``datetime.timezone.utc`` so the ORM receives a tz-aware value and
+       no shift occurs.
+
+    Pass ``timezone=True`` when constructing the field (done automatically
+    by :meth:`sqladmin.forms.ModelConverter.conv_datetime` for columns
+    declared with ``DateTime(timezone=True)``).
+    """
+
+    def __init__(self, *args: Any, timezone: bool = False, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._timezone = timezone
+
+    def process_data(self, value: Any) -> None:
+        """Convert tz-aware datetime to naive UTC before display."""
+        if value is not None and hasattr(value, "tzinfo") and value.tzinfo is not None:
+            # Normalise to UTC, then strip tzinfo so WTForms renders it
+            # without any offset suffix.  The HTML datetime-local input
+            # does not carry timezone information, so UTC is the safest
+            # unambiguous representation.
+            value = value.astimezone(dt_timezone.utc).replace(tzinfo=None)
+        super().process_data(value)
+
+    def process_formdata(self, valuelist: list[str]) -> None:
+        """Parse submitted value and re-attach UTC when timezone=True."""
+        super().process_formdata(valuelist)
+        if self._timezone and self.data is not None:
+            if self.data.tzinfo is None:
+                self.data = self.data.replace(tzinfo=dt_timezone.utc)
 
 
 class IntervalField(fields.StringField):
--- a/sqladmin/forms.py
+++ b/sqladmin/forms.py
@@ -49,6 +49,7 @@ from sqladmin.fields import (
     BooleanField,
     DateField,
     DateTimeField,
+    DateTimeLocalField,
     FileField,
     IntervalField,
     JSONField,
@@ -411,6 +412,14 @@ class ModelConverter(ModelConverterBase):
         prop: ColumnProperty,
         kwargs: dict[str, Any],
     ) -> UnboundField:
+        column = prop.columns[0]
+        # Use DateTimeLocalField for timezone-aware columns so that the
+        # UTC offset is not silently dropped when pre-populating the edit
+        # form, which would cause the stored time to shift on save.
+        # See https://github.com/aminalaee/sqladmin/issues/796
+        if getattr(column.type, "timezone", False):
+            kwargs["timezone"] = True
+            return DateTimeLocalField(**kwargs)
         return DateTimeField(**kwargs)
 
     @converts("Enum")


sqladmin/fields.py
```diff
@@ -59,6 +61,56 @@ class DateTimeField(fields.DateTimeField):
     widget = sqladmin_widgets.DateTimePickerWidget()  # type: ignore[assignment]
 
 
+class DateTimeLocalField(DateTimeField):
```
mmzeynalli, line 64: Lets go with `TimezoneAwareDateTimeField`, which is more descriptive.

sqladmin/forms.py
```diff
@@ -411,6 +412,14 @@ def conv_datetime(
...
         kwargs: dict[str, Any],
     ) -> UnboundField:
+        column = prop.columns[0]
+        # Use DateTimeLocalField for timezone-aware columns so that the
+        # UTC offset is not silently dropped when pre-populating the edit
+        # form, which would cause the stored time to shift on save.
+        # See https://github.com/aminalaee/sqladmin/issues/796
+        if getattr(column.type, "timezone", False):
+            kwargs["timezone"] = True
+            return DateTimeLocalField(**kwargs)
```
mmzeynalli, line 422: Non-Postgres backends with timezone=True. MySQL ignores the flag and returns naive datetimes; with this change, saves now send tz-aware values, and PyMySQL renders them as ... +00:00, which MySQL only accepts from 8.0.19. I think we need to mention this in docs, or handle it better.

sqladmin/forms.py
mmzeynalli, line 405: We need timezone aware time as well

sqladmin/fields.py
```diff
@@ -59,6 +61,56 @@ class DateTimeField(fields.DateTimeField):
...
+    declared with ``DateTime(timezone=True)``).
+    """
+
+    def __init__(self, *args: Any, timezone: bool = False, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._timezone = timezone
+
+    def process_data(self, value: Any) -> None:
+        """Convert tz-aware datetime to naive UTC before display."""
+        if value is not None and hasattr(value, "tzinfo") and value.tzinfo is not None:
```
mmzeynalli, line 98: if value is not None and getattr(value, "tzinfo", None)

mmzeynalli (changes requested):
Tests are also missing.

Vansh-Sharma27:
I checked this at `dd17bde`. The core fix works — I confirmed the round trip preserves the instant, which is the thing the issue is actually about:

```
stored in DB      : 2026-07-26 15:00:00+05:30
shown in edit form: 2026-07-26 09:30:00
submitted unchanged -> 2026-07-26 09:30:00+00:00
same instant?     : True
shift             : 0:00:00
```

So the reported time-shift is genuinely gone. Three things I ran into beyond the missing tests.

### 1. The displayed wall-clock changes for non-UTC data

This is deliberate and the docstring explains the reasoning, but it's a visible behaviour change worth calling out explicitly: a row stored as `15:00+05:30` previously rendered as `15:00` in the edit form and now renders as `09:30`. The instant is preserved on save either way, but someone who entered "3pm" will now open the form and see "9:30am".

For a field backed by `DateTimePickerWidget` that users interact with directly, that may surprise people even though it's more correct. Might be worth a line in the PR description so it isn't discovered after release.

### 2. `process_data` and `process_formdata` are asymmetric when `timezone=False`

`process_data` strips tzinfo for any aware value regardless of `self._timezone`, but `process_formdata` only re-attaches UTC when `self._timezone` is true. Constructing the field directly with the default gives a silent conversion:

```
timezone=False (default):
  stored   : 2026-07-26 15:00:00+05:30
  displayed: 2026-07-26 09:30:00
  submitted: 2026-07-26 09:30:00      <- naive, tz dropped
  tz lost? : True
```

`conv_datetime` always passes `timezone=True`, so nothing in the current codebase hits this. But `DateTimeLocalField` is added to `__all__`, which makes it public API, and the asymmetry is the kind of thing that bites later. Guarding `process_data` with `self._timezone` too would make the two halves symmetric.

### 3. The name collides with an existing WTForms field

WTForms already ships `DateTimeLocalField`:

```
wtforms  -> <class 'wtforms.fields.datetime.DateTimeLocalField'>
   widget: DateTimeLocalInput     (renders <input type="datetime-local">)
sqladmin -> <class 'sqladmin.fields.DateTimeLocalField'>
   bases: ['DateTimeField'], widget: DateTimePickerWidget
```

Two different things with the same name, and since `sqladmin/fields.py` already does `from wtforms import fields` and subclasses `fields.DateTimeField` elsewhere, a reader could reasonably assume this one wraps the WTForms field. Something like `TimezoneAwareDateTimeField` would describe what it actually does — it's about tz handling, not about the `datetime-local` input type.

### On the missing tests

Rather than just agreeing, here's the shape I used to verify the fix — it fails on `main` and passes here, so it should work as a regression test:

```python
def test_timezone_aware_datetime_roundtrip():
    ist = timezone(timedelta(hours=5, minutes=30))
    stored = datetime(2026, 7, 26, 15, 0, tzinfo=ist)

    class F(Form):
        dt = DateTimeLocalField(timezone=True)

    rendered = F(data={"dt": stored}).dt._value()
    submitted = F(DummyData(dt=rendered)).dt.data

    assert submitted == stored          # same instant, no shift
    assert submitted.tzinfo is not None  # still aware
```

`DummyData` is already in `tests/common.py`. Worth a second case asserting the create path too, since `conv_datetime` now returns the new field for both create and edit forms.

### What I ran

- the round-trip and `timezone=False` probes above, at `dd17bde`
- `pytest tests/` on this branch — 483 passed, 4 skipped; no regressions (the count matches `main`'s passing suite — the +1 seen on sibling PRs is their added regression test, which is exactly what's missing here)

mmzeynalli:
@AdarshJ173 any updates?

mmzeynalli:
Closing due to inactivity. This PR cannot be changed by maintainers, so new PR (#1142) was opened.

PR #1098, opened 2026-07-20
feat: support validation error handling during model saving

horpto:
This PR enables the application to catch and display errors raised within asynchronous `on_model_change` hooks. Previously, exceptions raised in an async context during the creation or editing of models might not have been properly caught and presented to the user. Now, `ValidationError`s raised during asynchronous lifecycle hooks are correctly captured and rendered in the corresponding form fields.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -721,6 +721,13 @@ class Admin(BaseAdminView):
         form_data_dict = self._denormalize_wtform_data(form.data, model_view.model)
         try:
             obj = await model_view.insert_model(request, form_data_dict)
+
+        except ValidationError as exc:
+            exc.enrich_form(form)
+
+            return await self.templates.TemplateResponse(
+                request, model_view.create_template, context, status_code=400
+            )
         except Exception as e:
             logger.exception(e)
             context["error"] = str(e)
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -40,6 +40,7 @@ from sqladmin._types import ENGINE_TYPE, SESSION_MAKER
 from sqladmin.ajax import QueryAjaxModelLoader
 from sqladmin.authentication import AuthenticationBackend, login_required
 from sqladmin.editors import collect_form_media
+from sqladmin.exceptions import ValidationError
 from sqladmin.flash import get_flashed_messages
 from sqladmin.forms import WTFORMS_ATTRS, WTFORMS_ATTRS_REVERSED
 from sqladmin.helpers import (
@@ -783,6 +784,12 @@ class Admin(BaseAdminView):
                 obj = await model_view.update_model(
                     request, pk=request.path_params["pk"], data=form_data_dict
                 )
+        except ValidationError as exc:
+            exc.enrich_form(form)
+
+            return await self.templates.TemplateResponse(
+                request, model_view.edit_template, context, status_code=400
+            )
         except Exception as e:
             logger.exception(e)
             context["error"] = str(e)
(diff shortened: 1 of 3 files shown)


Vansh-Sharma27 (commented):
I went through this one at `d05b6ae` since I've worked in this area before. The field-error path works well, but I ran into three cases where the behaviour doesn't hold. Reproductions below.

Setup: worktree at `d05b6ae`, `uv sync --all-extras`, wtforms 3.2.2, Python 3.11.

### 1. Form-level errors are never displayed

`ValidationError("some message")` stores positional args in `self.form_errors`, and `enrich_form` appends them to `form.form_errors`. But `form_errors` isn't read by any template:

```
$ grep -rn "form_errors" --include=*.html --include=*.py .
./sqladmin/exceptions.py:19,20,27,28,31
```

`exceptions.py` is the only hit. `_macros.html` renders `field.errors` (lines 69 and 74), and `create.html` / `edit.html` render `{{ error }}` from the context at line 14 — neither one touches `form.form_errors`.

So this returns a 400 with the form re-rendered and no message anywhere on the page:

```python
async def on_model_change(self, data, model, is_created, request):
    if data.get("name") == "formlevel":
        raise ValidationError("Cross-field check failed")
```

```
[form-level] status=400
[form-level] message present in body: False
```

### 2. The default-error fallback hits the same dead end

A bare `ValidationError()` takes the `elif not self.field_errors` branch and appends `"Validation error"` to `form.form_errors`:

```
[bare] status=400
[bare] 'Validation error' in body: False
```

That branch reads like it exists to guarantee the user always sees *something*, so I think this is the one worth fixing first.

The smallest fix that reuses what's already there is probably to put form-level messages into `context["error"]`, which both `create.html:14` and `edit.html:14` already render, instead of adding a new template block. Setting it next to the existing `exc.enrich_form(form)` call would keep the change local to the two handlers.

### 3. An unknown field name returns a 500

`enrich_form` indexes `form._fields[field_name]` directly, so a name that isn't on the form raises `KeyError`. It is raised *inside* the `except ValidationError` handler, so the `except Exception` below it doesn't catch it:

```
sqladmin\application.py:726: in create
    exc.enrich_form(form)
...
E           KeyError: 'does_not_exist'
sqladmin\exceptions.py:25: KeyError
```

A typo in a field name turning into a 500 seems worth guarding against. Separately, WTForms already exposes this publicly — `Form.__getitem__` (`wtforms/form.py:62`) just returns `self._fields[name]` — so `form[field_name]` works here without reaching for the private attribute.

### One thing the PR undersells

The title and the test say "asynchronous", but `Admin.create` and `Admin.edit` are the same code path for both engine types, and `on_model_change` is `async def` in the base class regardless. This works on sync engines too — I checked with the same view on `sync_engine`:

```
[sync] status=400
[sync] field message present: True
```

A sync test and a title without "asynchronous" would reflect the change better.

### What I ran

- `uv run pytest tests/test_views/test_view_async.py::test_async_validation_error` — passes
- `uv run pytest tests/` — 484 passed, 4 skipped; no regressions from this branch
- `uv run mypy sqladmin/exceptions.py sqladmin/application.py` — clean
- `uv run ruff check sqladmin/exceptions.py sqladmin/application.py` — clean

The field-error case works and the API is pleasant to use. It's really just the form-level path that needs somewhere to render.

mmzeynalli:
@Vansh-Sharma27 nice review! Some additions

1. /edit does not have tests
2. Catch scope includes after_model_change, which runs post-commit — so a ValidationError raised there would show a form error after the row was already saved. Edge case, not worth blocking on, but a one-line "only meaningful in on_model_change" note in the docstring would prevent surprise

mmzeynalli:
@horpto any updates?

horpto:
Sorry guys, I'm on vacation. I'll be back next week or the week after.

PR #1101, opened 2026-07-23
[Feature] Added `where` option in `form_ajax_refs`

### LINKED ISSUE #801: Adding custom filtering for relationships ModelConverterBase._prepare_select_options
### Checklist

I want to filter relationships in a query, but I can't because the ModelConverterBase._prepare_select_options function generates the query in a hardcode. 

```python
class CustomModelConverter(ModelConverter):
    select_option_stmts = {
        'field': select(Model).where(Model.level > 1)
    }


class ModelBaseConverter:
    select_option_stmts: Dict[str, Select]
    
    async def _prepare_select_options(
        self,
        prop: RelationshipProperty,
        session_maker: sessionmaker,
    ) -> list[tuple[str, Any]]:
        target_model = prop.mapper.class_
        if prop.key in self.select_option_stmts:
            stmt = self.select_option_stmts[prop.key]
        else:
            stmt = select(target_model)

        if is_async_session_maker(session_maker):
            async with session_maker() as session:
                objects = await session.execute(stmt)
                return [
                    (str(self._get_identifier_value(obj)), str(obj))
                    for obj in objects.scalars().unique().all()
                ]
        else:
                objects = await anyio.to_thread.run_sync(session.execute, stmt)
                return [
                    (str(self._get_identifier_value(obj)), str(obj))
                    for obj in objects.scalars().unique().all()
                ]


class View(ModelView, model=Model):
    form_converter = CustomModelConverter
```

 There is an option to inherit from ModelConverter and override the _prepare_select_options method in CustomModelConverter. But it looks like a huge crutch.
 
 ```python
 class CustomModelConverter(ModelConverter):
     async def _prepare_select_options(
        self,
        prop: RelationshipProperty,
        session_maker: sessionmaker,
    ) -> list[tuple[str, Any]]:
        target_model = prop.mapper.class_
        stmt = <custom query>

        if is_async_session_maker(session_maker):
            async with session_maker() as session:
                objects = await session.execute(stmt)
                return [
                    (str(self._get_identifier_value(obj)), str(obj))
                    for obj in objects.scalars().unique().all()
                ]
        else:
                objects = await anyio.to_thread.run_sync(session.execute, stmt)
                return [
                    (str(self._get_identifier_value(obj)), str(obj))
                    for obj in objects.scalars().unique().all()
                ]
                
                
class View(ModelView, model=Model):
    form_converter = CustomModelConverter
 ```

### Additional context

_No response_

maxim-f1:
## Added `where` option to `form_ajax_refs` for AJAX filtering

`form_ajax_refs` now supports a where clause to filter AJAX lookup results.
The expression is passed directly to SQLAlchemy – full flexibility, but no validation at config time. 
```
form_ajax_refs = {
    "user": {
        "fields": ("id", "name"),
        "order_by": "id",
        "where": (User.active.is_(True),),
    }
}
```
Screenshot from 2026-07-23 08-53-20
Screenshot from 2026-07-23 08-53-32
Closes #801


sqladmin/ajax.py
```diff
@@ -59,6 +64,57 @@ def _process_fields(self) -> list:
...
+
+        if isinstance(self.order_by, (str, InstrumentedAttribute)):
+            self.order_by = [self.order_by]
+        elif not isinstance(self.order_by, Iterable):
+            raise ValueError(
+                f"The form_ajax_refs.field.order_by field accepts only str and "
+                f"sqlalchemy.orm.attributes.InstrumentedAttribute "
+                f"or collections of them. "
+                f"Received: {self.order_by}"
+            )
```
mmzeynalli: The order_by "stricter validation" is a backward-incompatible. 

The old code passed whatever you gave it straight into stmt.order_by(o), so order_by=(Team.id.desc(),), desc("id"), or func.lower(Team.name) all worked. The new _process_order_by_fields accepts only str or InstrumentedAttribute and raises ValueError at add_view time for anything else. Team.id.desc() is a UnaryExpression, not an InstrumentedAttribute → existing configs that sort descending now crash on startup. For a PR whose headline is where, silently removing descending/expression order_by is a trap. It should keep accepting arbitrary order-by expressions (validate loosely, e.g. accept ColumnElement/str and resolve strings only).

sqladmin/ajax.py
```diff
@@ -59,6 +64,57 @@ def _process_fields(self) -> list:
...
+
+        for field in self.order_by:
+            if isinstance(field, str):
+                attr = getattr(self.model, field, None)
+
+            elif isinstance(field, InstrumentedAttribute):
+                attr = getattr(self.model, field.key, None)
```
mmzeynalli: Even when you pass a valid InstrumentedAttribute, the code discards it and re-looks-it-up on self.model by .key. If the attribute belongs to a joined/related model (a legitimate old use), getattr(self.model, other_key) is None or the wrong column → "does not exist" or a silently incorrect sort. field is already usable — use it directly instead of round-tripping through field.key.

docs/cookbook/optimize_relationship_loading.md
```diff
@@ -42,6 +42,7 @@ class ParentAdmin(ModelView, model=Parent):
         "children": {
             "fields": ("id",),
             "order_by": "id",
+            "where": (Children.id > 1,),
```
mmzeynalli: Add more docs

docs/configurations.md
```diff
@@ -421,6 +421,7 @@ The forms are based on `WTForms` package and include the following options:
             "address": {
                 "fields": ("zip_code", "street"),
                 "order_by": ("id",),
+                "where": (Address.active.is_(True),),
```
mmzeynalli: SAme here

sqladmin/ajax.py
```diff
@@ -59,6 +64,57 @@ def _process_fields(self) -> list:
...
+            self.where = [self.where]
+
+        if isinstance(self.where, Iterable) is False or any(
+            not isinstance(item, ColumnElement) for item in self.where
+        ):
+            raise ValueError(
+                f'The "where" field only accepts SQLAlchemy ColumnElement '
+                f"or an iterable of SQLAlchemy ColumnElement expressions. "
+                f"Got: {self.where}. "
+                f'Example: "where": (and_(User.name == "example"),)'
```
mmzeynalli: and_ is extra

tests/test_ajax.py
```diff
@@ -443,3 +493,194 @@ async def test_format_by_pk_returns_empty_for_missing_record() -> None:
...
+
+async def test_error_on_where_validation_in_ajax() -> None:
+    class MissedFieldAdmin(ModelView, model=MissedField):
+        form_ajax_refs = {"team": {"fields": ("id",), "where": ("id = 1",)}}
+
+    error_msg = (
+        'The "where" field only accepts SQLAlchemy ColumnElement or an iterable of '
+        "SQLAlchemy ColumnElement expressions. "
+        "Got: ('id = 1',). "
+        'Example: "where": (and_(User.name == "example"),)'
```
mmzeynalli: Same here

tests/test_ajax.py
```diff
@@ -443,3 +493,194 @@ async def test_format_by_pk_returns_empty_for_missing_record() -> None:
...
+
+async def test_not_iterable_on_where_validation_in_ajax() -> None:
+    class MissedFieldAdmin(ModelView, model=MissedField):
+        form_ajax_refs = {"team": {"fields": ("id",), "where": "id = 1"}}
+
+    error_msg = (
+        'The "where" field only accepts SQLAlchemy ColumnElement or an iterable of '
+        "SQLAlchemy ColumnElement expressions. "
+        "Got: id = 1. "
+        'Example: "where": (and_(User.name == "example"),)'
```
mmzeynalli: Same here

sqladmin/ajax.py
```diff
@@ -29,7 +31,8 @@ def __init__(
         self.model = model
         self.model_admin = model_admin
         self.fields = options.get("fields", {})
-        self.order_by = options.get("order_by")
+        self.where = options.get("where", [])
+        self.order_by = options.get("order_by", [])
```
mmzeynalli: It would be really nice, if we could make it dynamic with request. For example, there are 1 admin and 1 superadmin. Superadmin can access all entries that are active, admin can access entires that are active and status = 'READY'.

mmzeynalli: Right now it is static and "decided" in build time

sqladmin/ajax.py
```diff
@@ -59,6 +64,57 @@ def _process_fields(self) -> list:
 
         return remote_fields
 
+    def _process_where_conditions(self) -> list[ColumnElement]:
+        if isinstance(self.where, ColumnElement):
+            self.where = [self.where]
```
mmzeynalli: self.where can be tuple, so your return type is might be incorrect

mmzeynalli (changes requested):

sqladmin/ajax.py
maxim-f1: Implementing this functionality will require significant modifications, since `QueryAjaxModelLoader` is typically invoked at the `widget` level, where the request object is not passed through.

sqladmin/ajax.py
maxim-f1: Fix validation for edge cases and add tests.

sqladmin/ajax.py
maxim-f1: Fix validation for edge cases and add tests.

sqladmin/ajax.py
maxim-f1: fixed

tests/test_ajax.py
```diff
@@ -443,3 +493,244 @@ async def test_format_by_pk_returns_empty_for_missing_record() -> None:
...
+        form_ajax_refs = {
+            "team": {
+                "fields": (MissedField.id,),
+                "order_by": func.lower(MissedField.team_id),
+            }
+        }
+
+    admin.add_view(MissedFieldAdmin)
+
+    assert len(MissedFieldAdmin()._form_ajax_refs["team"]._cached_fields_order_by) == 1
```
mmzeynalli, line 909: Also check that get_list returns correct data

sqladmin/ajax.py
mmzeynalli: application.py:924, you can pass request to get_list function.

mmzeynalli (changes requested):

tests/test_ajax.py
maxim-f1: There is the same variable there, and the tests check that variable

sqladmin/ajax.py
maxim-f1: Passing the `request` to `get_list` will not allow customizing the logic because `QueryAjaxModelLoader` is not designed for that. To add custom logic, you need to override the entire class and override the method.

mmzeynalli: Not if you can pass:

```
def where_function(request):
  if request.session.is_superadmin:
    return User.status.in_('active', 'pending')

  return User.status == 'acitve'

form_ajax_refs = {
    "user": {
        "fields": ("id", "name"),
        "order_by": "id",
        "where": where_function,
    }
}
```

maxim-f1: Completely reworked the entire concept.

mmzeynalli: I still think class is extensive. Just adding support for a function would have been enough. Also, support both function and ColumntElement, right now you have dropped latter.

docs/cookbook/optimize_relationship_loading.md
```diff
@@ -46,6 +46,52 @@ class ParentAdmin(ModelView, model=Parent):
...
+  from starlette.requests import Request
+  from sqlalchemy.sql.elements import ColumnElement
+  
+  async def where_clause(request: Request, term: str) -> ColumnElement:
+      return Children.id == 1
+  ```
+
+
+
+
```
mmzeynalli: Remove these extra lines

sqladmin/ajax.py
```diff
@@ -59,6 +72,90 @@ def _process_fields(self) -> list:
...
+
+        if callable(self.where):
+            clause = self.where(request, term)
+
+            if isinstance(clause, Awaitable):
+                clause = await clause  # type: ignore[union-attr]
+
+            if not isinstance(clause, ColumnElement):
+                raise ValueError(
+                    f'Function "{type(self.where).__name__}" '
```
mmzeynalli: self.where.__name__ would print the actual function name

This one returns: `Function "function" in "where" option should return ColumnElement`

sqladmin/ajax.py
```diff
@@ -59,6 +72,90 @@ def _process_fields(self) -> list:
...
+                    f'in "where" option should return ColumnElement. '
+                    f"Got: {type(clause).__name__}"
+                )
+
+            return [clause]
```
mmzeynalli: considering we can do

'where': [clause1, clause2], it would be logical to let function to be able to reutrn list of clauses as well

sqladmin/ajax.py
```diff
@@ -59,6 +72,90 @@ def _process_fields(self) -> list:
 
         return remote_fields
 
+    def _check_where_clauses_valid(self) -> bool:
```
mmzeynalli: Isnt this the smae as _prepare_where_clauses?

mmzeynalli (changes requested):
Nice work! Some little pesky comments left)

sqladmin/ajax.py
maxim-f1: We can't fully check what to use the entered valid data (because of the functions that we can check only in runtime), but we can check at least the type hints. Thus, the user will learn about the error as soon as possible, even before the search starts

sqladmin/ajax.py
maxim-f1: In theory, this can be done, but it will further complicate the already large type hint

sqladmin/ajax.py
mmzeynalli: I think we can let this go. It is users responsibility to configure it right with correct config with given documentation. We cannot be 100% failsafe

mmzeynalli (approved):

PR #1102, opened 2026-07-23
[Feature] Added `check_can_create`

maxim-f1:
## Added `check_can_create`
Example: Only one record can be created in the table.
```
async def check_can_create(self, request) -> bool:
    async with self.session_maker() as session:
        result = await session.execute(select(User))
        return len(result.all()) < 1
```
Screenshot from 2026-07-23 11-14-49
Screenshot from 2026-07-23 11-14-57

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -153,7 +153,7 @@ class BaseAdmin:
             if isinstance(view, ModelView) and view.identity == identity:
                 return view
 
-        raise HTTPException(status_code=404)
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
 
     def add_view(self, view: type[ModelView] | type[BaseView]) -> None:
         """Add ModelView or BaseView classes to Admin.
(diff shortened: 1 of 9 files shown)


tests/test_views/test_view_async.py
```diff
@@ -744,6 +755,21 @@ async def test_check_can_view_details(client: AsyncClient) -> None:
...
+    admin.add_view(ActionAdmin)
+
+    response = await client.get("admin/action/list")
+
+    assert 'href="http://testserver/admin/action/create"' not in response.text
+
+    response = await client.post("/admin/action/create", data={})
+    assert response.status_code == 403
+
+
```
mmzeynalli, line 775: assert GET /admin/action/create → 403

sqladmin/application.py
```diff
@@ -321,7 +322,12 @@ async def _list(self, request: Request) -> None:
     async def _create(self, request: Request) -> None:
         model_view = self._find_model_view(request.path_params["identity"])
         if not model_view.can_create or not model_view.is_accessible(request):
-            raise HTTPException(status_code=403)
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
+
+        if hasattr(model_view, "check_can_create"):
```
mmzeynalli: This is always True. I believe there are other parts with similar hasattr, remove them as well

sqladmin/application.py
```diff
@@ -321,7 +322,12 @@ async def _list(self, request: Request) -> None:
...
+            can_create = await model_view.check_can_create(request)
+            if can_create is not True:
```
mmzeynalli: Lets be "permissive" here. IF user accidentally returns 1/0 here, this would fail, so lets do `if not can_create`

sqladmin/application.py
```diff
@@ -321,7 +322,12 @@ async def _list(self, request: Request) -> None:
     async def _create(self, request: Request) -> None:
         model_view = self._find_model_view(request.path_params["identity"])
         if not model_view.can_create or not model_view.is_accessible(request):
```
mmzeynalli, line 324: If I have can_create False, and check_can_create True for specific person, it would still fail here. Consider it

sqladmin/application.py
```diff
@@ -321,7 +322,12 @@ async def _list(self, request: Request) -> None:
...
-            raise HTTPException(status_code=403)
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
+
+        if hasattr(model_view, "check_can_create"):
+            can_create = await model_view.check_can_create(request)
+            if can_create is not True:
+                raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
```
mmzeynalli: Update all other statuses to starlette status

mmzeynalli (changes requested):

sqladmin/application.py
maxim-f1: That is the behavior that was intended. The general `can_create` setting acts as a master switch for everyone.

tests/test_views/test_view_async.py
maxim-f1: fixed

sqladmin/application.py
maxim-f1: fixed

sqladmin/application.py
maxim-f1: fixed

sqladmin/application.py
maxim-f1: done

sqladmin/application.py
mmzeynalli: Okay, then just mention it on the docs. The rest is fine

maxim-f1: Done

mmzeynalli (approved):

PR #1112, opened 2026-07-31
fix(pretty_export): stringify related objects when joining list values

### LINKED ISSUE #1111: `PrettyExport` with `relationship` and `_base_export_cell` return error
### Checklist

### Describe the bug

If you put relationship in PrettyExport, an error will occur.

You can reproduce it through a test:
```
async def test_get_export_row_values_basic(self):
        column_list = ["id", "name", "email", "addresses"]
        session_maker = session_maker
        is_async = False

        address = Address(id=1, user_id=1)
        user = User(id=1, name="John Doe", email="john@example.com", is_active=True)
        session.add(user)
        session.add(address)
        session.commit()
        user = session.get(User, 1)

        model_view = UserAdmin()
        column_names = ["id", "name", "email", "addresses"]

        values = await PrettyExport._get_export_row_values(
            model_view, user, column_names
        )

    assert len(values) == 3
    assert values[0] == 1
    assert values[1] == "John Doe"
    assert values[2] == "john@example.com"
```

### Expected behavior

Formatted value

### Actual behavior

```
if isinstance(value, list):
  cell_value = ",".join(formatted_value)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 0: expected str instance, Address found
sqladmin/pretty_export.py:29: TypeError
```

### Debugging material

_No response_

### Environment

sqladmin 0.30.0

### Additional context

_No response_

Sanjays2402:
Closes #1111

`PrettyExport._base_export_cell` joined a relationship's formatted value with `",".join(formatted_value)`, but for an `InstrumentedList` the default formatter returns the related model instances themselves, so exporting a `column_list` containing a to-many relationship raised `TypeError: sequence item 0: expected str instance, Address found`. Each item is now coerced with `str()` before joining, matching how the non-list relationship branch renders its value.

Regression test added in `tests/test_pretty_export.py` beside the existing `_get_export_row_values` tests; it fails with the reported TypeError without the fix and passes with it.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -1,4 +1,4 @@
-from collections.abc import AsyncGenerator, Iterable
+from collections.abc import AsyncGenerator
 from typing import TYPE_CHECKING, Any
 
 from starlette.responses import StreamingResponse
@@ -15,7 +15,7 @@ class PrettyExport:
     @staticmethod
     async def _base_export_cell(
         model_view: "ModelView", name: str, value: Any, formatted_value: Any
-    ) -> str:
+    ) -> Any:
         """
         Default formatting logic for a cell in pretty export.
 
@@ -27,7 +27,10 @@ class PrettyExport:
         if name in model_view._relation_names:
             if isinstance(formatted_value, str):
                 cell_value = formatted_value
-            elif isinstance(formatted_value, Iterable):
+            elif isinstance(formatted_value, (set, frozenset)):
+                # Unordered collections are sorted so exports are reproducible.
+                cell_value = ",".join(sorted(str(item) for item in formatted_value))
+            elif isinstance(formatted_value, (list, tuple)):
                 cell_value = ",".join(str(item) for item in formatted_value)
             else:
                 cell_value = formatted_value
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -30,6 +30,8 @@ class PrettyExport:
             elif isinstance(formatted_value, (set, frozenset)):
                 # Unordered collections are sorted so exports are reproducible.
                 cell_value = ",".join(sorted(str(item) for item in formatted_value))
+            elif isinstance(formatted_value, dict):
+                cell_value = ",".join(str(item) for item in formatted_value.values())
             elif isinstance(formatted_value, (list, tuple)):
                 cell_value = ",".join(str(item) for item in formatted_value)
             else:
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -1,4 +1,4 @@
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Iterable
 from typing import TYPE_CHECKING, Any
 
 from starlette.responses import StreamingResponse
@@ -25,7 +25,9 @@ class PrettyExport:
         Only used when `use_pretty_export = True`.
         """
         if name in model_view._relation_names:
-            if isinstance(value, list):
+            if isinstance(formatted_value, str):
+                cell_value = formatted_value
+            elif isinstance(formatted_value, Iterable):
                 cell_value = ",".join(str(item) for item in formatted_value)
             else:
                 cell_value = formatted_value
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -26,7 +26,7 @@ class PrettyExport:
         """
         if name in model_view._relation_names:
             if isinstance(value, list):
-                cell_value = ",".join(formatted_value)
+                cell_value = ",".join(str(item) for item in formatted_value)
             else:
                 cell_value = formatted_value
         else:
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -35,7 +35,9 @@ class PrettyExport:
             elif isinstance(formatted_value, (list, tuple)):
                 cell_value = ",".join(str(item) for item in formatted_value)
             else:
-                cell_value = formatted_value
+                # A to-one relationship: honour __str__ like every other branch,
+                # but keep None so it still exports as an empty cell.
+                cell_value = formatted_value if formatted_value is None else str(formatted_value)
         else:
             if isinstance(value, bool):
                 cell_value = "TRUE" if value else "FALSE"
--- a/sqladmin/pretty_export.py
+++ b/sqladmin/pretty_export.py
@@ -37,7 +37,7 @@ class PrettyExport:
             else:
                 # A to-one relationship: honour __str__ like every other branch,
                 # but keep None so it still exports as an empty cell.
-                cell_value = formatted_value if formatted_value is None else str(formatted_value)
+                cell_value = None if formatted_value is None else str(formatted_value)
         else:
             if isinstance(value, bool):
                 cell_value = "TRUE" if value else "FALSE"
--- a/tests/test_pretty_export.py
+++ b/tests/test_pretty_export.py
@@ -7,7 +7,12 @@ from typing import Any
 
 import pytest
 from sqlalchemy import Boolean, Column, ForeignKey, Integer, String
-from sqlalchemy.orm import declarative_base, relationship, sessionmaker
+from sqlalchemy.orm import (
+    attribute_keyed_dict,
+    declarative_base,
+    relationship,
+    sessionmaker,
+)
 from starlette.responses import StreamingResponse
 
 from sqladmin import ModelView
@@ -30,6 +35,9 @@ class User(Base):
 
     addresses = relationship("Address", back_populates="user")
     tags = relationship("Tag", back_populates="user", collection_class=set)
+    items = relationship(
+        "Item", back_populates="user", collection_class=attribute_keyed_dict("sku")
+    )
 
 
 class Address(Base):
@@ -57,6 +65,19 @@ class Tag(Base):
 
     def __str__(self) -> str:
         return self.label
+
+
+class Item(Base):
+    __tablename__ = "items"
+
+    id = Column(Integer, primary_key=True)
+    sku = Column(String)
+    user_id = Column(Integer, ForeignKey("users.id"))
+
+    user = relationship("User", back_populates="items")
+
+    def __str__(self) -> str:
+        return self.sku
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


sqladmin/pretty_export.py
```diff
@@ -26,7 +26,7 @@ async def _base_export_cell(
         """
         if name in model_view._relation_names:
             if isinstance(value, list):
-                cell_value = ",".join(formatted_value)
+                cell_value = ",".join(str(item) for item in formatted_value)
```
mmzeynalli, line 36: With this, if the user has a custom formatter on a relationship column, it gets character-split. If someone sets `column_formatters = {"addresses": lambda m, a: f"{len(m.addresses)} addresses"}`, value is still an `InstrumentedList` so the `isinstance(value, list)` branch is taken, but `formatted_value` is now a string — and joining a string iterates its characters:

id,addresses
1,"2, ,a,d,d,r,e,s,s,e,s"     # expected: 2 addresses

Another problem, if relationship is set backed, then this if will skip:

With collection_class=set, value isn't a list, so it falls through to cell_value = formatted_value and the raw set repr lands in the CSV — using each element's repr, so a __str__ on the model doesn't help:

1,{<myapp.models.Tag object at 0x7f10349cd090>}

This is not your problem, but still fix it pls

tests/test_pretty_export.py
```diff
@@ -125,6 +125,32 @@ class UserAdmin(ModelView, model=User):
...
+            column_names = ["id", "name", "addresses"]
+
+            values = await PrettyExport._get_export_row_values(
+                model_view, user, column_names
+            )
+
+            assert len(values) == 3
+            assert values[0] == 1
+            assert values[1] == "John Doe"
+            assert values[2] == ",".join(str(address) for address in user.addresses)
```
mmzeynalli: Add `__str__` to Adress

mmzeynalli (changes requested):

mmzeynalli:
One more thing. This PR considers relationships rendered as QuerySelectField (scalar — uselist=False, a plain many-to-one). It does not fix relationships rendered as QuerySelectMultipleField. 

```
column_import_list = [User.name, User.addresses]
CSV:  name,addresses
      Bob,adg34gfb13

branch result: {"ok": true, "imported": 1, "skipped": 0, "missed_rows": []}
```

This PR works by keeping errors that fallback_form.validate() already reported. It depends on WTForms actually reporting one. For a collection relationship, `validate()` returns `True` with `errors == {}` even though the value is garbage. There is no error to keep, so the new block at lines 130–135 has nothing to copy and falls through silently.

sqladmin/pretty_export.py
Sanjays2402: Both real, fixed in 9a3e5a6.

The branch was on `isinstance(value, list)`, so it dispatched on the raw attribute and then joined whatever the formatter returned. Now it dispatches on `formatted_value`: `str` passes through untouched, any other `Iterable` gets `",".join(str(item) ...)`, everything else falls through. That covers the string-formatter case and `collection_class=set` (and any other collection class) in one branch instead of enumerating types.

Added `__str__` to `Address`, a `Tag` relationship with `collection_class=set`, and two tests: the set case and the `column_formatters` case returning a plain string. Both fail on the previous commit with exactly what you described (`2, ,a,d,d,r` for the formatter one). 18 passed, ruff clean.

Your last comment about `column_import_list` / `fallback_form.validate()` looks like it was meant for #1120 — this PR only touches the export path. You are right that the collection case slips through there since WTForms returns `errors == {}` for a garbage `QuerySelectMultipleField` value, so there is nothing for the keep-errors block to copy. I will handle that over on #1120 rather than widen this one.

sqladmin/pretty_export.py
```diff
@@ -25,8 +25,10 @@ async def _base_export_cell(
         Only used when `use_pretty_export = True`.
         """
         if name in model_view._relation_names:
-            if isinstance(value, list):
-                cell_value = ",".join(formatted_value)
+            if isinstance(formatted_value, str):
+                cell_value = formatted_value
+            elif isinstance(formatted_value, Iterable):
```
mmzeynalli: (list, set, tuple, frozenset)

Lets do this instead of Iterable. Apparently, formatted_value can be dict as well:

```py
class Order(Base):
    items = relationship("Item", collection_class=attribute_keyed_dict("sku"))
```

mmzeynalli: Even, do like this:

```py
if isinstance(value, bool):
    cell_value = "TRUE" if value else "FALSE"
elif isinstance(formatted_value, str):
    cell_value = formatted_value
elif isinstance(formatted_value, (set, frozenset)):
    cell_value = ",".join(sorted(str(item) for item in formatted_value))
elif isinstance(formatted_value, (list, tuple)):
    cell_value = ",".join(str(item) for item in formatted_value)
else:
    cell_value = formatted_value
```

Bcuz, set and frozenset are unodered. and each export might give different set of items.

sqladmin/pretty_export.py
```diff
@@ -25,8 +25,10 @@ async def _base_export_cell(
...
+                cell_value = ",".join(str(item) for item in formatted_value)
             else:
                 cell_value = formatted_value
```
mmzeynalli, line 38: Add test for this as well, please

sqladmin/pretty_export.py
mmzeynalli: return type is Any

mmzeynalli (changes requested):

sqladmin/pretty_export.py
Sanjays2402: Applied your ladder in d4b0691, with the explicit types instead of `Iterable`.

```py
if isinstance(formatted_value, str):
    cell_value = formatted_value
elif isinstance(formatted_value, (set, frozenset)):
    cell_value = ",".join(sorted(str(item) for item in formatted_value))
elif isinstance(formatted_value, (list, tuple)):
    cell_value = ",".join(str(item) for item in formatted_value)
else:
    cell_value = formatted_value
```

I kept the `bool` check in the non-relationship branch where it already was rather than hoisting it above the relationship branch - a bool can't be a relationship value, so moving it only changes which branch a non-relationship bool takes, and it currently takes that one.

Two tests added: `test_get_export_row_values_with_set_relationship_is_sorted` (three tags inserted as vip/beta/alpha, asserts `alpha,beta,vip`, so it pins the ordering rather than just the membership) and `test_get_export_row_values_with_dict_relationship` using `collection_class=attribute_keyed_dict("sku")`. Both fail on the previous commit - the dict one used to come out as `sku-1` because `Iterable` caught the dict and joined its keys.

One thing to decide: with the explicit tuple, an `attribute_keyed_dict` relationship now falls through to `else` and the raw dict reaches the writer, so the CSV cell is the dict repr. That is what your snippet does and it is at least not silently wrong the way key-joining was, but if you'd rather it render as `",".join(sorted(str(v) for v in formatted_value.values()))` say so and I'll add the branch plus flip the test.

Also changed the `_base_export_cell` return annotation from `str` to `Any` - it was never accurate, the `else` arms return whatever the formatter gave (int, date, None), and now a dict too.

sqladmin/pretty_export.py
mmzeynalli: Keyed collections (`attribute_keyed_dict`) aren't list/tuple/set/str, so they
fall through here and reach the CSV as a Python repr:

    1,"Main St,Second St","alpha,vip","{'sku-1': <__main__.Item object at 0x7f55d58e03d0>, ...}"

Two runs of the same export over the same rows:

    run 1: ...<__main__.Item object at 0x7f55d58e03d0>...
    run 2: ...<__main__.Item object at 0x7fe63fae0390>...

Which is the opposite of the comment three lines up — sets became reproducible in
the same commit that left mappings not reproducible. It also leaks the module path
and ignores `__str__` on the related model, which every other branch honours.

9a3e5a6 had this right via `Iterable` (it produced `sku-1,sku-2`). Narrowing away
from `Iterable` was a sound instinct, but could we name `dict` rather than let it
fall through?

    elif isinstance(formatted_value, dict):
        cell_value = ",".join(str(item) for item in formatted_value.values())

mmzeynalli (changes requested):

sqladmin/pretty_export.py
mmzeynalli, line 41: Test missing for this line.

What still lands here is a to-one relationship, and on a related model without
`__str__` that exports as the same thing this commit just fixed for mappings:

    column_export_list = ["id", "author", "publisher"]
      Author has no __str__  ->  <__main__.Author object at 0x7f...>
      Publisher has __str__  ->  P

    CSV LINE -> 1,<__main__.Author object at 0x7f...>,P

Identical on current main, so not something this PR caused — but it's the same bug class,
so we might fix it in this PR

mmzeynalli (changes requested):
Perfect. Almost ready! We just need one more test)

Sanjays2402:
Thanks @mmzeynalli — that last test is in (ed1f827).

The to-one fall-through now honours `__str__` like every other branch: `cell_value = None if formatted_value is None else str(formatted_value)`, so a related model with `__str__` exports its string form and a missing relation still exports as an empty cell (not the string "None").

Added `test_get_export_row_values_with_to_one_relationship` with `Book`/`Publisher` models — Publisher defines `__str__` returning its name, so the test pins `publisher` exporting as "P", plus the None case. Also verified no line exceeds 88 chars for ruff.

sqladmin/pretty_export.py
```diff
@@ -25,10 +25,19 @@ async def _base_export_cell(
...
+                # Unordered collections are sorted so exports are reproducible.
+                cell_value = ",".join(sorted(str(item) for item in formatted_value))
+            elif isinstance(formatted_value, dict):
+                cell_value = ",".join(str(item) for item in formatted_value.values())
+            elif isinstance(formatted_value, (list, tuple)):
+                cell_value = ",".join(str(item) for item in formatted_value)
+            else:
+                # A to-one relationship: honour __str__ like every other branch,
+                # but keep None so it still exports as an empty cell.
+                cell_value = None if formatted_value is None else str(formatted_value)
```
mmzeynalli, line 44: `make lint` fails this. Kindly ask you to add install pre-commit to this repo, considering you are frequently contributing

sqladmin/pretty_export.py
```diff
@@ -25,10 +25,19 @@ async def _base_export_cell(
...
                 cell_value = formatted_value
+            elif isinstance(formatted_value, (set, frozenset)):
+                # Unordered collections are sorted so exports are reproducible.
+                cell_value = ",".join(sorted(str(item) for item in formatted_value))
+            elif isinstance(formatted_value, dict):
+                cell_value = ",".join(str(item) for item in formatted_value.values())
+            elif isinstance(formatted_value, (list, tuple)):
+                cell_value = ",".join(str(item) for item in formatted_value)
+            else:
+                # A to-one relationship: honour __str__ like every other branch,
```
mmzeynalli: Small thing, but this comment describes a mechanism that isn't running.

A missing to-one relation never arrives here as None. `BASE_FORMATTERS` maps
`type(None)` to `empty_formatter` (formatters.py:80), which returns `Markup("")`
— a str subclass — so it's caught by the `isinstance(formatted_value, str)`
branch on line 28 and never reaches this one:

```
    publisher present   formatted=<Pub object at 0x7f...>  branch=ELSE  cell='P'
    publisher MISSING   formatted=Markup('')               branch=str   cell=Markup('')
```

That's why 01a93b6 had to change the test to assert "" — which is right, the
behaviour is correct, it just isn't for the reason stated here.

So, maybe:

```
    # A to-one relationship: str() it so this branch matches the others. A
    # custom column_formatters entry may return None; keep that as None rather
    # than exporting the string "None". A *missing* relation does not arrive
    # here -- BASE_FORMATTERS turns it into Markup("") and the str branch takes it.
```

sqladmin/pretty_export.py
mmzeynalli: Please resolve comments, or comment "resolved", so I know this is done

mmzeynalli (commented):
Almost done, nothing serious left.

sqladmin/pretty_export.py
Sanjays2402: Resolved — that was a mypy error (`Incompatible types in assignment`, `str | None` vs `str`); added a `cell_value: Any` annotation. `ruff check`, `ruff format --check` and `mypy` all pass now. And noted — I run pre-commit locally and will keep `make lint` green on my PRs; leaving it out of this PR's scope as suggested.

sqladmin/pretty_export.py
Sanjays2402: Resolved — replaced with your suggested wording. I rewrapped it so the lines fit the repo's 88-char limit; the wording itself is unchanged.

sqladmin/pretty_export.py
Sanjays2402: Resolved — all three threads addressed in the latest commit (2cd2708). Thanks for the close read!

Sanjays2402:
Thanks for the careful review! Addressed the latest round:

- **Lint** — `make lint` failure was a mypy error (`Incompatible types in assignment` at the to-one branch); added a `cell_value: Any` annotation. `ruff check`, `ruff format --check` and `mypy` are all green now. And point taken on pre-commit — I have it installed and run it locally, so I'll keep `make lint` clean on my PRs; left it out of this PR's scope.
- **Stale comment** — replaced with your suggested wording (just rewrapped to fit the repo's 88-char line limit).

Also replied on each thread so you can mark them resolved. Tests pass (23/23 in `tests/test_pretty_export.py`). Let me know if anything else needs a look!

mmzeynalli (approved):
Nice, thanks!

PR #1038, opened 2026-04-23
Create form_details_query

MaximDementyev:
Fix #1037 
The form_details_query method is implemented
The difference from form_edit_query in the loop line (in form_edit_query self._form_relations variable used)

Issue #1037, opened 2026-04-23
Incorrect ModelView.details_query method


MaximDementyev:
### Checklist

- [x] The bug is reproducible against the latest release or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

The details_query method uses form_edit_query for some reason
https://github.com/smithyhq/sqladmin/blob/984b7458c10fccd1379f8bc876e985bcc4461634/sqladmin/models.py#L1226-L1232

This causes an error when used in back_populates in relation when using hybrid_property

### Steps to reproduce the bug

```
from fastapi import FastAPI
from sqladmin import Admin, ModelView
from sqlalchemy import Column, ForeignKey, Integer, String, create_engine, select
from sqlalchemy.ext.hybrid import hybrid_property
from sqlalchemy.orm import declarative_base, relationship, selectinload

Base = declarative_base()
engine = create_engine("sqlite:///example.db",connect_args={"check_same_thread": False})

class Component(Base):
    __tablename__ = "component"

    id = Column(Integer, primary_key=True)
    name = Column(String)

    def __str__(self):
        return self.name

class Product(Base):
    __tablename__ = "products"

    id = Column(Integer, primary_key=True)
    name = Column(String)
    product_component = relationship('ProductComponent', back_populates='product')

    def __str__(self):
        return self.name

class ProductComponent(Base):
    __tablename__ = "product_component"

    id = Column(Integer, primary_key=True)

    product_id = Column(Integer, ForeignKey('products.id'))
    component_id = Column(Integer, ForeignKey('component.id'))

    product = relationship(Product,  back_populates='product_component', lazy='immediate')
    component = relationship(Component, lazy='immediate')

    @hybrid_property
    def product_name(self):
        return self.product.name

    @product_name.inplace.expression
    def _product_name_expression(cls):
        return select(Product.name).where(Product.id == cls.product_id).label('product_name')

    @hybrid_property
    def component_name(self):
        return self.component.name

    @product_name.inplace.expression
    def _component_name_expression(cls):
        return select(Component.name).where(Component.id == cls.component_id).label('component_name')

    def __str__(self):
        return f'{self.product_name}_{self.component_name}'

Base.metadata.create_all(engine)

class ProductAdmin(ModelView, model=Product):
    column_list = [Product.name]
    form_columns = [Product.name]

class ComponentAdmin(ModelView, model=Component):
    column_list = [Component.name]
    form_columns = [Component.name]

[...]

PR #1113, opened 2026-07-31
Increase coverage

### LINKED ISSUE #1111: `PrettyExport` with `relationship` and `_base_export_cell` return error

maxim-f1:
Closes #1111


tests/test_views/test_view_async.py
maxim-f1, line 141: @mmzeynalli 
This is a very strange function. Coverage shows it's not used. I don't really understand the scenarios in which it's called, or even how to call it.

maxim-f1:
@mmzeynalli 
https://github.com/smithyhq/sqladmin/blob/64372dd623672e42c715600eeb072a9a5cab8535/tests/test_base_view.py#L30
Can you tell me in more detail what needs to be tested in this case?

tests/test_views/test_view_async.py
mmzeynalli: It is by @MaximDementyev, and absolutely have no idea what this does

mmzeynalli:
> @mmzeynalli
> 
> https://github.com/smithyhq/sqladmin/blob/64372dd623672e42c715600eeb072a9a5cab8535/tests/test_base_view.py#L30
> 
> 
> Can you tell me in more detail what needs to be tested in this case?

Before identity of baseview was alphabetically, right now it is first function by order. So if I have

```python
class View(BaseView):
  def b1(): ...
  def a1(): ...
  def c1(): ...
```

Before, `a1` was identity (alphabetical), now it is `b1` (order in which they are defined)

tests/test_views/test_view_async.py
MaximDementyev: The test was added at [your request.](https://github.com/smithyhq/sqladmin/pull/1038#issuecomment-4496325623)
This test is a simplified version of the code that was described in issue https://github.com/smithyhq/sqladmin/issues/1037

sqladmin/pretty_export.py
```diff
@@ -26,14 +26,15 @@ async def _base_export_cell(
         """
         if name in model_view._relation_names:
             if isinstance(value, list):
-                cell_value = ",".join(formatted_value)
+                cell_value = ",".join([str(value) for value in formatted_value])
             else:
-                cell_value = formatted_value
+                cell_value = str(formatted_value)
```
mmzeynalli, line 31: Undo this, it is being worked on rn on #1111

sqladmin/_import.py
```diff
@@ -180,7 +180,7 @@ async def foreign_key_error_message(
 ) -> str | None:
     try:
         coerced_value = coerce_column_value(target_column, value)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError):  # pragma: no cover
         return f"Invalid value {value!r} for column {column_key}."
```
mmzeynalli: No need to cover this one, as it is reachable. Add tests instead

tests/test_authentication.py
```diff
@@ -282,3 +279,69 @@ def test_extra_session_kwargs_passed_to_middleware() -> None:
...
+        form = await request.form()
+        if form.get("username") != "success":
+            return False
+
+        request.session.update({"token": "valid_token"})
+        return True
+
+    async def logout(self, request: Request) -> bool:  # pragma: no cover
+        request.session.clear()
+        return True
```
mmzeynalli: This is already covered isnt it?
In `test_authenticate_func_always_return_bool`

sqladmin/_types.py
```diff
@@ -24,7 +24,7 @@
 
 if sys.version_info < (3, 11):
 
-    class StrEnum(str, Enum):
+    class StrEnum(str, Enum):  # pragma: no cover
```
mmzeynalli, line 28: No need for this, as we have Github runners for Python 3.10, which covers this

tests/test_views/test_view_async.py
```diff
@@ -1317,6 +1324,101 @@ async def check_can_import(self, request: Request) -> bool:
...
 
+@pytest.mark.parametrize(
+    "call_number, expected_text",
+    [
+        (1, '"total": 1, "imported": 0'),
+        (2, '"total": 1, "imported": 0'),
+        (3, "Import canceled. No rows were imported"),
+    ],
+)
+async def test_import_csv_reqeust_disconnect(
```
mmzeynalli: reqeust -> request

tests/test_views/test_view_sync.py
```diff
@@ -1232,6 +1253,120 @@ async def check_can_import(self, request: Request) -> bool:
...
 
 
+@pytest.mark.parametrize(
+    "call_number, expected_text",
+    [
+        (1, '"total": 1, "imported": 0'),
+        (2, '"total": 1, "imported": 0'),
+        (3, "Import canceled. No rows were imported"),
+    ],
+)
```
mmzeynalli, line 1280: I added a fourth parametrize case (99, '"total": 1, "imported": 0') — a call number so high the disconnect never fires — and got 4 passed

tests/test_views/test_view_async.py
```diff
@@ -1363,6 +1470,58 @@ async def test_import_csv_invalid_extension(client: AsyncClient) -> None:
...
+        raise Exception("Error!")
+
+    original_func = sqladmin.helpers.parse_csv
+
+    for module_name, module in list(sys.modules.items()):
+        if (
+            hasattr(module, "parse_csv")
+            and getattr(module, "parse_csv") is original_func
+        ):
+            monkeypatch.setattr(module, "parse_csv", mock_parse_csv)
```
mmzeynalli: monkeypatch.setattr("sqladmin._import.parse_csv", mock_parse_csv)

Why not?

tests/test_views/test_view_async.py
```diff
@@ -1331,6 +1433,11 @@ async def test_import_csv_bad_type_is_404(client: AsyncClient) -> None:
     assert response.status_code == 404
 
 
+async def test_import_csv_empty_payload_error(client: AsyncClient) -> None:
+    response = await client.post("/admin/user/import")
+    assert "Invalid file upload. Expected a CSV file." in response.text
```
mmzeynalli, line 1455: check response status as well, always

tests/test_views/test_view_sync.py
```diff
@@ -1144,6 +1150,21 @@ def test_hybrid_property(client: TestClient) -> None:
...
+        person = Person(name="Daniel")
+        session.add(person)
+        session.flush()
+        worker = Worker(person_id=person.id)
+        session.add(worker)
+        session.commit()
+
+    response = client.get("/admin/worker/details/1")
+    assert response.status_code == 200
+    assert "Hybrid" in response.text
```
mmzeynalli, line 1168: I think this one is misnamed. The details page reads person_name off an instance, so it runs the Python getter; _person_name_expression is only invoked by class-level access, which happens in WorkerAdmin.column_details_list at class-definition time rather than in anything the test asserts. As written it's close to a duplicate of test_hybrid_property at :1142.

The "Hybrid" suffix on the getter's return is a good idea for telling the two paths apart — could we lean on it? Something that goes through the list/query path and asserts the value came from the SQL expression (no "Hybrid" suffix) would actually pin the inplace branch. Same for test_view_async.py:2045

tests/test_fields.py
```diff
@@ -231,8 +231,35 @@ class FRequired(Form):
 
 def test_textarea_field() -> None:
     class F(Form):
-        text = TextAreaField()
+        text = TextAreaField(validators=[DataRequired()])
```
mmzeynalli: could the validators=[DataRequired()] change go into a new fourth test rather than into this one? Right now nothing constructs a bare TextAreaField() anymore, and that's the default configuration a user gets without passing anything.

tests/test_pretty_export.py
```diff
@@ -41,6 +44,9 @@ class Address(Base):
 
     user = relationship("User", back_populates="addresses")
 
+    def __str__(self) -> str:
+        return f"{self.id}"
```
mmzeynalli: Make thiss Address {self.id}

tests/test_views/test_view_async.py
mmzeynalli: Also add id to parameters, to know which test fails if only one of them is failing

mmzeynalli (changes requested):

sqladmin/pretty_export.py
maxim-f1: I'll merge the final result into this branch. I'll leave it for now so that new tests for this case don't fail.

sqladmin/_types.py
maxim-f1: Versions that don't fall under this if clause report this line as missing (which is actually true). Therefore, add "No cover"

tests/test_views/test_view_sync.py
maxim-f1: In theory, this is the correct behavior; the call_counter counter has not reached call_number, which means that the mock is_disconnected will return False

tests/test_views/test_view_sync.py
maxim-f1: Maybe we should remove the _person_name_expression function entirely? Sqladmin doesn't call this code directly anywhere, and there are no cases for it either. And testing sqlalchemy's capabilities isn't our goal.

sqladmin/_import.py
maxim-f1: Fixed

tests/test_authentication.py
maxim-f1: Fixed

tests/test_views/test_view_async.py
maxim-f1: Fixed

tests/test_views/test_view_async.py
maxim-f1: Fixed

tests/test_fields.py
maxim-f1: Fixed

tests/test_pretty_export.py
maxim-f1: Fixed

tests/test_fields.py
```diff
@@ -236,3 +236,41 @@ class F(Form):
...
+
+
+def test_textarea_field_all_bool_false() -> None:
+    class F(Form):
+        text = TextAreaField(enable_autoresize=False, show_chars_count=False)
+
+    form = F()
+    assert "autoresize-textarea" not in form.text()
+    assert "chars-count-label" not in form.text()
+
```
mmzeynalli, line 275: TextArea.__call__ is never invoked with flags set, so line widgets.py:187 — `kwargs[k] = getattr(flags, k)`, the body of the validation_attrs loop — is no longer reached.

tests/test_application.py
```diff
@@ -360,3 +364,83 @@ def test_is_list_template_global():
...
+            logo_url: str | None = None,
+            logo_width: int = 64,
+            logo_height: int = 64,
+            favicon_url: str | None = None,
+            middlewares: Sequence[Middleware] | None = None,
+            debug: bool = False,
+            templates_dir: str = "templates",
+            authentication_backend: AuthenticationBackend | None = None,
+            static_files_kwargs: dict[str, Any] | None = None,
+            i18n_config: I18nConfig | None = None,
```
mmzeynalli, line 389: add audit_backend as well as param

tests/test_import.py
```diff
@@ -111,3 +130,41 @@ async def test_validate_import_row_reports_coercion_errors() -> None:
...
+    monkeypatch.setattr(
+        "sqladmin._import.coerce_column_value", mock_coerce_column_value
+    )
+
+    model_view = _model_view(ImportWidgetAdmin)
+    result = await validate_foreign_key_values(
+        model_view,
+        {
+            ImportWidget.id.key: 1,
+            ImportWidget.active.key: True,
```
mmzeynalli, line 165: This is neither coerced, nor used?

tests/test_pretty_export.py
```diff
@@ -61,22 +67,30 @@ async def _get_csv_content(self, response: StreamingResponse) -> str:
 
     async def test_get_export_row_values_basic(self):
         class UserAdmin(ModelView, model=User):
-            column_list = ["id", "name", "email"]
+            column_list = ["id", "name", "email", "addresses"]
             session_maker = session_maker
             is_async = False
```
mmzeynalli, line 72: remove this line, it is autodetected from the session_maker

tests/test_pretty_export.py
mmzeynalli, line 374: there are no such fields, how di we miss this? Could you also fix this test?

mmzeynalli, line 365: also keep 1 of them, why 2?

mmzeynalli (changes requested):

PR #1120, opened 2026-08-02
fix(import): report invalid relationship values during CSV import

### LINKED ISSUE #1114: Import from csv with parameter `continue_on_error: 0` ignore relationship errors
### Checklist

### Describe the bug

If you pass the `relationship` variable with incorrect parameters into a CSV file for import, and set `continue_on_error: 0`, then the record will be saved as valid but without the relationship.

```py
async def test_import_csv_continue_on_error_false(client: AsyncClient) -> None:
    class AddressContinueeOnErrorFalseAdmin(ModelView, model=Address):
        can_import = True
        column_import_list = [Address.id, Address.user]

    local_app = Starlette()
    local_admin = Admin(app=local_app, engine=engine)
    local_admin.add_view(AddressContinueeOnErrorFalseAdmin)

    transport = ASGITransport(app=local_app)
    async with AsyncClient(
        transport=transport,
        base_url="http://continue_on_error",
    ) as local_client:
        
        response = await local_client.post(
            "/admin/address/import",
            data={"continue_on_error": "no"},
            files={
                "csvfile": (
                    "address.csv",
                    b"id,user\r\n1,adg34gfb13\r\n",
                    "text/csv",
                )
            },
        )
        assert 'Import aborted on invalid row 2. No rows were imported' in response.text
```

### Expected behavior

Stop import and output error.

### Actual behavior

Import continues as if nothing happened.

### Debugging material

Studying the code led me to these two function calls. It is assumed that the error should have been caught in one of them:

https://github.com/smithyhq/sqladmin/blob/64372dd623672e42c715600eeb072a9a5cab8535/sqladmin/_import.py#L443

https://github.com/smithyhq/sqladmin/blob/64372dd623672e42c715600eeb072a9a5cab8535/sqladmin/_import.py#L451

Here is what comes when you run the test from the example:
```py
print(f'{row = }')
processed += 1
merged_import_data, row_errors, row_data = validate_import_row(
    row,
    model_view._import_prop_names,
    model_view.model,
    form_class,
    denormalize_wtform_data,
)
print(f'{merged_import_data = }')

foreign_key_errors = await validate_foreign_key_values(
    model_view=model_view,
    row_data=merged_import_data,
    fk_error_cache=fk_error_cache,
)
for field_name, field_errors in foreign_key_errors.items():
    row_errors.setdefault(field_name, []).extend(field_errors)

print(f'{row_errors = }')
```
```py
row = MultiDict([('id', '1'), ('user', 'adg34gfb13')])
merged_import_data = {'id': 1, 'user': None}
row_errors = {}
```

### Environment

sqladmin 0.30.0

### Additional context

_No response_

Sanjays2402:
Closes #1114

Relationship fields are not mapper columns, so `merge_import_row_data()` neither coerces nor re-validates them. WTForms does reject an unresolvable value with `Not a valid choice`, but it denormalizes it to `None`, and the second validation pass then sees an empty (valid) field — so the row imported as valid with the relationship silently dropped and `continue_on_error=False` never aborted the import.

`validate_import_row()` now keeps the errors the fallback form already reported for import columns that are not mapper columns. Regression test added in `tests/test_import.py`; it fails on `main` (no `profile` error) and passes with the fix.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/sqladmin/_import.py
+++ b/sqladmin/_import.py
@@ -109,8 +109,15 @@ def validate_import_row(
     """Coerce CSV values, then run WTForms validation on coerced form input."""
     row_data = {col: row.get(col) for col in import_columns}
 
+    # Relationship fields are not mapper columns, so they are not coerced or
+    # re-validated below: an unresolvable value is denormalized to None and would
+    # silently pass. Only those fields need the extra validation pass, so skip it
+    # entirely when every import column is a mapper column.
+    mapper_columns = sa_inspect(model).columns
+    non_mapper_columns = [col for col in import_columns if col not in mapper_columns]
+
     fallback_form = form_class(row)
-    fallback_valid = fallback_form.validate()
+    fallback_valid = fallback_form.validate() if non_mapper_columns else True
     fallback_data = denormalize_wtform_data(fallback_form.data, model)
 
     merged_import_data, row_errors = merge_import_row_data(
@@ -121,12 +128,10 @@ def validate_import_row(
     )
 
     if not fallback_valid:
-        # Relationship fields are not mapper columns, so they are not coerced or
-        # re-validated below: an unresolvable value is denormalized to None and
-        # would silently pass. Keep the errors WTForms already reported for them.
-        mapper_columns = sa_inspect(model).columns
-        for field_name, field_errors in fallback_form.errors.items():
-            if field_name in import_columns and field_name not in mapper_columns:
+        # Keep the errors WTForms already reported for the non-mapper columns.
+        for field_name in non_mapper_columns:
+            field_errors = fallback_form.errors.get(field_name)
+            if field_errors:
                 row_errors.setdefault(field_name, []).extend(field_errors)
 
     validation_form = form_class(
--- a/sqladmin/_import.py
+++ b/sqladmin/_import.py
@@ -130,6 +130,10 @@ def validate_import_row(
     if not fallback_valid:
         # Keep the errors WTForms already reported for the non-mapper columns.
         for field_name in non_mapper_columns:
+            # An empty cell is not a bad value: the pass below already reports it
+            # when the field is required, so do not double up here.
+            if row_data.get(field_name) in (None, ""):
+                continue
             field_errors = fallback_form.errors.get(field_name)
             if field_errors:
                 row_errors.setdefault(field_name, []).extend(field_errors)
@@ -139,7 +143,8 @@ def validate_import_row(
     )
     if not validation_form.validate():
         for field_name, field_errors in validation_form.errors.items():
-            row_errors.setdefault(field_name, []).extend(field_errors)
+            existing = row_errors.setdefault(field_name, [])
+            existing.extend(e for e in field_errors if e not in existing)
 
     return merged_import_data, row_errors, row_data
 
--- a/sqladmin/_import.py
+++ b/sqladmin/_import.py
@@ -113,12 +113,19 @@ def validate_import_row(
     # re-validated below: an unresolvable value is denormalized to None and would
     # silently pass. Only those fields need the extra validation pass, so skip it
     # entirely when every import column is a mapper column.
-    mapper_columns = sa_inspect(model).columns
-    non_mapper_columns = [col for col in import_columns if col not in mapper_columns]
+    mapper = sa_inspect(model)
+    non_mapper_columns = [
+        col for col in import_columns if mapper.columns.get(col) is None
+    ]
 
-    fallback_form = form_class(row)
-    fallback_valid = fallback_form.validate() if non_mapper_columns else True
-    fallback_data = denormalize_wtform_data(fallback_form.data, model)
+    if non_mapper_columns:
+        fallback_form = form_class(row)
+        fallback_valid = fallback_form.validate()
+        fallback_data = denormalize_wtform_data(fallback_form.data, model)
+    else:
+        fallback_form = None
+        fallback_valid = True
+        fallback_data = {}
 
     merged_import_data, row_errors = merge_import_row_data(
         model,
@@ -129,6 +136,7 @@ def validate_import_row(
 
     if not fallback_valid:
         # Keep the errors WTForms already reported for the non-mapper columns.
+        assert fallback_form is not None
         for field_name in non_mapper_columns:
             # An empty cell is not a bad value: the pass below already reports it
             # when the field is required, so do not double up here.
@@ -144,7 +152,11 @@ def validate_import_row(
     if not validation_form.validate():
         for field_name, field_errors in validation_form.errors.items():
             existing = row_errors.setdefault(field_name, [])
-            existing.extend(e for e in field_errors if e not in existing)
+            seen = set(existing)
+            for error in field_errors:
+                if error not in seen:
+                    existing.append(error)
+                    seen.add(error)
 
     return merged_import_data, row_errors, row_data
 
--- a/sqladmin/_import.py
+++ b/sqladmin/_import.py
@@ -118,14 +118,9 @@ def validate_import_row(
         col for col in import_columns if mapper.columns.get(col) is None
     ]
 
-    if non_mapper_columns:
-        fallback_form = form_class(row)
-        fallback_valid = fallback_form.validate()
-        fallback_data = denormalize_wtform_data(fallback_form.data, model)
-    else:
-        fallback_form = None
-        fallback_valid = True
-        fallback_data = {}
+    fallback_form = form_class(row)
+    fallback_valid = fallback_form.validate() if non_mapper_columns else True
+    fallback_data = denormalize_wtform_data(fallback_form.data, model)
 
     merged_import_data, row_errors = merge_import_row_data(
         model,
@@ -136,7 +131,6 @@ def validate_import_row(
 
     if not fallback_valid:
         # Keep the errors WTForms already reported for the non-mapper columns.
-        assert fallback_form is not None
         for field_name in non_mapper_columns:
             # An empty cell is not a bad value: the pass below already reports it
             # when the field is required, so do not double up here.
--- a/sqladmin/_import.py
+++ b/sqladmin/_import.py
@@ -110,6 +110,7 @@ def validate_import_row(
     row_data = {col: row.get(col) for col in import_columns}
 
     fallback_form = form_class(row)
+    fallback_valid = fallback_form.validate()
     fallback_data = denormalize_wtform_data(fallback_form.data, model)
 
     merged_import_data, row_errors = merge_import_row_data(
@@ -118,6 +119,15 @@ def validate_import_row(
         row_data,
         fallback_data,
     )
+
+    if not fallback_valid:
+        # Relationship fields are not mapper columns, so they are not coerced or
+        # re-validated below: an unresolvable value is denormalized to None and
+        # would silently pass. Keep the errors WTForms already reported for them.
+        mapper_columns = sa_inspect(model).columns
+        for field_name, field_errors in fallback_form.errors.items():
+            if field_name in import_columns and field_name not in mapper_columns:
+                row_errors.setdefault(field_name, []).extend(field_errors)
 
     validation_form = form_class(
         build_import_form_row(row, merged_import_data, import_columns)
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -2,7 +2,7 @@ import enum
 
 import pytest
 from sqlalchemy import Boolean, Column, Enum, ForeignKey, Integer, String
-from sqlalchemy.orm import declarative_base, sessionmaker
+from sqlalchemy.orm import declarative_base, relationship, sessionmaker
 from starlette.datastructures import MultiDict
 
 from sqladmin import ModelView
@@ -34,6 +34,8 @@ class ImportWidget(Base):
     profile_id = Column(Integer, ForeignKey("import_profile_validate.id"))
     active = Column(Boolean, nullable=False)
 
+    profile = relationship("ImportProfile")
+
 
 class ImportProfile(Base):
     __tablename__ = "import_profile_validate"
@@ -47,6 +49,13 @@ class ImportUserAdmin(ModelView, model=ImportUser):
 
 class ImportWidgetAdmin(ModelView, model=ImportWidget):
     column_import_list = [ImportWidget.profile_id, ImportWidget.active]
+
+
+@pytest.fixture(autouse=True)
+def prepare_database():
+    Base.metadata.create_all(engine)
+    yield
+    Base.metadata.drop_all(engine)
 
 
 def _model_view(admin_class: type[ModelView]) -> ModelView:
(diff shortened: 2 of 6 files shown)


sqladmin/_import.py
```diff
@@ -110,6 +110,7 @@ def validate_import_row(
     row_data = {col: row.get(col) for col in import_columns}
 
     fallback_form = form_class(row)
+    fallback_valid = fallback_form.validate()
     fallback_data = denormalize_wtform_data(fallback_form.data, model)
```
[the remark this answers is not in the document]
Sanjays2402: Good catch, fixed in aa41844.

I compute the non-mapper import columns up front and only call `fallback_form.validate()` when that list is non-empty, so an import whose columns are all mapper columns does exactly one validation pass per row as before. The error-copying loop now iterates that same list instead of scanning `fallback_form.errors`.

`tests/test_import.py` is 4 passed; the relationship test still fails if I force `fallback_valid = True`.

sqladmin/_import.py
```diff
@@ -119,6 +127,13 @@ def validate_import_row(
         fallback_data,
     )
 
+    if not fallback_valid:
+        # Keep the errors WTForms already reported for the non-mapper columns.
+        for field_name in non_mapper_columns:
+            field_errors = fallback_form.errors.get(field_name)
+            if field_errors:
+                row_errors.setdefault(field_name, []).extend(field_errors)
+
```
mmzeynalli, line 142: When the relationship's FK is non-nullable the field gets allow_blank=False, so the third pass at line 140 rejects the now-empty value as well and the same string gets appended twice:

```
[required rel, bad value] errors={'profile': ['Not a valid choice', 'Not a valid choice']}
[required rel, blank]     errors={'profile': ['Not a valid choice', 'Not a valid choice']}
```

mmzeynalli: Skipping empty cells here fixes it, since the third pass still reports genuinely-required-but-blank fields:

```python
if not fallback_valid:
        # An empty cell is not a bad value: the pass below already reports it
        # when the field is required, so do not double up here.
        for field_name in non_mapper_columns:
            if row_data.get(field_name) in (None, ""):
                continue
            field_errors = fallback_form.errors.get(field_name)
            if field_errors:
                row_errors.setdefault(field_name, []).extend(field_errors)
```

sqladmin/_import.py
mmzeynalli: We have to manage duplication here as well, snmth like this:

```py
for field_name, field_errors in validation_form.errors.items():
            existing = row_errors.setdefault(field_name, [])
            existing.extend(e for e in field_errors if e not in existing)
```

tests/test_import.py
```diff
@@ -111,3 +120,25 @@ async def test_validate_import_row_reports_coercion_errors() -> None:
...
+    merged, errors, _row_data = validate_import_row(
+        row,
+        model_view.get_import_columns(),
+        ImportWidget,
+        form_class,
+        Admin._denormalize_wtform_data,
+    )
+
+    assert merged["profile"] is None
+    assert "profile" in errors
```
mmzeynalli: assert errors["profile"] == ["Not a valid choice"]

Do this. to fix and prevent future error duplication

mmzeynalli: Add test for successful import as well

tests/test_import.py
```diff
@@ -49,6 +51,13 @@ class ImportWidgetAdmin(ModelView, model=ImportWidget):
     column_import_list = [ImportWidget.profile_id, ImportWidget.active]
 
 
+@pytest.fixture(autouse=True)
+def prepare_database():
+    Base.metadata.create_all(engine)
+    yield
+    Base.metadata.drop_all(engine)
```
mmzeynalli, line 69: why?

mmzeynalli (changes requested):

sqladmin/_import.py
Sanjays2402: All four in e608c12.

Double-reporting: the fallback copy now skips cells that are empty, so a required relationship with a blank value is only reported by the third pass, and the final pass extends with `e not in existing`. Both cases you pasted now come back as `errors={'profile': ['Not a valid choice']}`.

Test assertions are exact equality instead of `in` now, and I added `test_validate_import_row_does_not_duplicate_relationship_errors` parametrized over the bad value and the blank value against a new `ImportRequiredWidget` whose FK is non-nullable — that is the model that reproduces your `allow_blank=False` path. Both fail on the previous commit with "Left contains one more item: 'Not a valid choice'".

Also added `test_validate_import_row_accepts_valid_relationship_value` for the success path.

On the `prepare_database` fixture: `ImportProfile` and the widget tables were never created, so the relationship field's query for its choices hit a missing table. Without the fixture four of the tests in the file error out. It could move to `tests/common.py` if you'd rather it not be local.

`tests/test_import.py` is 7 passed, ruff check and format clean.

sqladmin/_import.py
```diff
@@ -109,7 +109,15 @@ def validate_import_row(
...
+    # Relationship fields are not mapper columns, so they are not coerced or
+    # re-validated below: an unresolvable value is denormalized to None and would
+    # silently pass. Only those fields need the extra validation pass, so skip it
+    # entirely when every import column is a mapper column.
+    mapper_columns = sa_inspect(model).columns
+    non_mapper_columns = [col for col in import_columns if col not in mapper_columns]
+
     fallback_form = form_class(row)
+    fallback_valid = fallback_form.validate() if non_mapper_columns else True
     fallback_data = denormalize_wtform_data(fallback_form.data, model)
```
mmzeynalli: When non_mapper_columns is empty (the common case — most models have no relationship in column_import_list), a full WTForms form is still instantiated and denormalized for every single row. For a 200k-row import this is a lot of wasted work.

Do smth like this:

```py
if non_mapper_columns:
    fallback_form = form_class(row)
    fallback_valid = fallback_form.validate()
    fallback_data = denormalize_wtform_data(fallback_form.data, model)
else:
    fallback_valid = True
    fallback_data = {}
```

tests/test_import.py
```diff
@@ -111,3 +131,72 @@ async def test_validate_import_row_reports_coercion_errors() -> None:
...
+        row,
+        model_view.get_import_columns(),
+        ImportWidget,
+        form_class,
+        Admin._denormalize_wtform_data,
+    )
+
+    assert errors == {}
+    assert merged["active"] is True
+    assert merged["profile"] == "1"
```
mmzeynalli, line 178: One little issue here:

merged["profile"] is "1" because QuerySelectField._data holds the pk as a string. When this gets downstream to _set_attributes_async, setattr(widget, "profile", "1") assigns the string "1" to a SQLAlchemy relationship attribute, which will either raise an AttributeError or silently fail.

sqladmin/_import.py
```diff
@@ -119,12 +127,24 @@ def validate_import_row(
...
+                row_errors.setdefault(field_name, []).extend(field_errors)
+
     validation_form = form_class(
         build_import_form_row(row, merged_import_data, import_columns)
     )
     if not validation_form.validate():
         for field_name, field_errors in validation_form.errors.items():
-            row_errors.setdefault(field_name, []).extend(field_errors)
+            existing = row_errors.setdefault(field_name, [])
+            existing.extend(e for e in field_errors if e not in existing)
```
mmzeynalli: `e not in existing` on a list is O(n) per element. For fields with many validation errors this quadratic scan is unnecessary. Lets use a set for the deduplication check.

tests/test_import.py
```diff
@@ -111,3 +131,72 @@ async def test_validate_import_row_reports_coercion_errors() -> None:
...
+
+
+class ImportRequiredWidgetRelationshipAdmin(ModelView, model=ImportRequiredWidget):
+    column_import_list = [ImportRequiredWidget.profile]
```
mmzeynalli, line 182: I think this should be profile_id

mmzeynalli (changes requested):
Add two tests:

1. Full streaming flow with invalid fkey: There is no test that sends a CSV with an invalid relationship value through import_csv → stream_import_response and asserts that the NDJSON stream reports the error correctly.

2. Invalid value for a nullable relationship field

`test_validate_import_row_does_not_duplicate_relationship_errors` uses a non-nullable case. But neither explicitly asserts the merged result when the nullable field has an invalid value — the test only checks errors. A row with an invalid nullable relationship should not be inserted as `profile = None`silently, but the merged assertion (`assert merged["profile"] is None`) is present but not paired with a check that the error actually stops the row from being imported.

Sanjays2402:
Pushed 65e2def.

- Avoids constructing and denormalizing the fallback form when all import fields are mapper columns.
- Uses a set-backed deduplication pass for validation errors.
- Adds a nullable relationship regression and an end-to-end `import_csv` → NDJSON streaming test proving an invalid relationship aborts the import with `profile: Not a valid choice`.

I kept the relationship field in the valid-value test rather than changing it to `profile_id`: `_set_attributes_async` already resolves relationship identifiers and writes the related FK, so `profile` is the intended intermediate import value. The focused import/async-view tests pass (11), Ruff check/format and mypy pass. The three relationship regressions fail against the pre-fix source and pass on this commit.

sqladmin/_import.py
```diff
@@ -109,8 +109,23 @@ def validate_import_row(
...
+    ]
+
+    if non_mapper_columns:
+        fallback_form = form_class(row)
+        fallback_valid = fallback_form.validate()
+        fallback_data = denormalize_wtform_data(fallback_form.data, model)
+    else:
+        fallback_form = None
+        fallback_valid = True
+        fallback_data = {}
```
mmzeynalli: `fallback_data` isn't only used for validation — it's passed to
`merge_import_row_data`, which reads it at `helpers.py:424-425` to fill in a
non-nullable column whose CSV cell is blank. With `{}` that branch can never
fire, so the column reaches the INSERT as NULL.

Repro on this branch — `active` is `Boolean(nullable=False)`, CSV is
`name,active` / `alpha,<blank>` / `beta,true`:

    MAIN 64372dd:  ok=True   rows: [(1,'alpha',False), (2,'beta',True)]
    THIS PR       65e2def:  ok=False  rows: []
      {"__all__": ["(sqlite3.IntegrityError) NOT NULL constraint failed: w.active
                    [SQL: INSERT INTO w (name, active) VALUES (?, ?)]
                    [parameters: ('alpha', None)]"]}

It only hits imports with *no* relationship column, which is the common case —
and is the shape of `ImportWidgetAdmin` above.

Only the validation needs gating, not the form itself:

    fallback_form = form_class(row)
    fallback_valid = fallback_form.validate() if non_mapper_columns else True
    fallback_data = denormalize_wtform_data(fallback_form.data, model)

Remove `assert` on line 139, which would be stripped under `python -O`.

Regression test that fails without the fix and passes with it:

    @pytest.mark.anyio
    async def test_validate_import_row_defaults_non_nullable_column_from_form():
        model_view = _model_view(ImportWidgetAdmin)
        form_class = await model_view.scaffold_form(model_view._form_create_rules)
        row = MultiDict([("profile_id", "1"), ("active", "")])

        merged, errors, _row_data = validate_import_row(
            row, model_view.get_import_columns(), ImportWidget,
            form_class, Admin._denormalize_wtform_data,
        )

        assert errors == {}
        assert merged == {"profile_id": 1, "active": False}

sqladmin/_import.py
```diff
@@ -109,8 +109,23 @@ def validate_import_row(
...
+    # Relationship fields are not mapper columns, so they are not coerced or
+    # re-validated below: an unresolvable value is denormalized to None and would
+    # silently pass. Only those fields need the extra validation pass, so skip it
+    # entirely when every import column is a mapper column.
+    mapper = sa_inspect(model)
+    non_mapper_columns = [
+        col for col in import_columns if mapper.columns.get(col) is None
+    ]
+
+    if non_mapper_columns:
```
mmzeynalli: This covers `QuerySelectField` but not `QuerySelectMultipleField`, so to-many
relationships still drop bad values silently. On this branch, `title,tags` /
`badpost,999`:

    ok=True  missed=[]   posts: [(1, 'badpost', [])]

The cause is in the field rather than here: `fields.py:306` reads
`self._invalid_formdata` before the `self.data` getter that sets it, so the
*first* `validate()` on a form always passes and only a second one catches it:

    first  validate() -> True   {}
    second validate() -> False  {'tags': ['Not a valid choice']}

Since `validate_import_row` builds a fresh form for each pass, the flag is never
observed. Moving the `self.data` read above the flag check would fix the field
and give this PR to-many coverage for free.

Happy for that to be a follow-up — but could we note it in the description? The
title says "relationship values" and right now half of them are still silent.

tests/test_import.py
```diff
@@ -111,3 +131,92 @@ async def test_validate_import_row_reports_coercion_errors() -> None:
...
+        ImportRequiredWidget,
+        form_class,
+        Admin._denormalize_wtform_data,
+    )
+
+    assert errors["profile"] == ["Not a valid choice"]
+
+
+@pytest.mark.anyio
+async def test_validate_import_row_rejects_invalid_nullable_relationship_value() -> (
```
mmzeynalli: This is the same test as `test_validate_import_row_reports_invalid_relationship_value`
on line 141 — same admin, same model, same row. The only difference is
`errors == {...}` vs `errors["profile"] == [...]`, and this one strictly
subsumes that one. Worth dropping one of them, or giving this a different input.

tests/test_views/test_view_async.py
```diff
@@ -299,6 +304,7 @@ class WithDefaultsAdmin(ModelView, model=WithDefaults):
 
 
 admin.add_view(UserAdmin)
+admin.add_view(UserRelationshipImportAdmin)
```
mmzeynalli: `test_import_csv_reports_invalid_relationship_value` builds its own `local_app` /
`local_admin`, so this line isn't needed — I removed it and all 65 tests in the
module still pass. What it does do is leave the shared `admin` holding two views
with identity `user` for every other test in the file. The rendered sidebar is
unaffected today, so this is tidiness rather than a bug, but shared-state edits
like this tend to bite later.

mmzeynalli (changes requested):

mmzeynalli (changes requested):
Perfect. One problem remains though, which is deeper problem not bcuz of this PR:

helpers.py:393 serialises the merged value with str(). For a multi-select the merged value is a list, so the second validation pass is handed the literal string "['1']":

```
row = MultiDict([("title","good"), ("tags","1")])      # Tag id=1 exists

pass 1: validate() = True    errors = {}
        form.data['tags']  = ['1']
        merged['tags']     = ['1']
rebuilt form row           = {'title': 'good', 'tags': "['1']"}     <-- str(list)
pass 2: validate() = False   errors = {'tags': ['Not a valid choice']}
```

That round-trip was always broken. It was invisible because QuerySelectMultipleField never reported unknown selections.

End to end:

```
CSV: title,tags / good,1      (Tag id=1 exists)
  merge-base 64372dd:  ok=True   tags: ['t1']
  prev head  65e2def:  ok=True   tags: ['t1']
  this head  b6492c1:  ok=False  {'tags': ['Not a valid choice']}
```

Sanjays2402:
Thanks @mmzeynalli — good spot, that's fixed in 3471d48.

`build_import_form_row` no longer `str()`-ifies a merged multi-select value. When the merged value is a list/tuple it now emits one form value per item (`form_row.append`), so the second validation pass receives exactly what the first one saw (`["1"]` instead of the literal string `"['1']"`). Your repro now behaves like main again: `title,tags` / `good,1` validates clean with `merged["tags"] == ["1"]`.

Added `test_validate_import_row_round_trips_valid_multi_select_value` with a new `ImportPost`/`ImportTag` one-to-many pair mirroring your repro — it fails on the previous commit with `{'tags': ['Not a valid choice']}`.

One thing to note (per your earlier ask to record it somewhere): an *invalid* to-many value is still silently dropped, because `QuerySelectMultipleField` only reports it on the second `validate()` of the same form and this PR builds a fresh form per pass. Happy to do the `fields.py:306` reorder as a follow-up — say the word.

mmzeynalli (approved):
Added two more tests. Nice work!

PR #1123, opened 2026-08-07
fix: fixed file selection widget

maxim-f1:
### An optional file can now be deselected even during creation. This is necessary in cases where the user has selected a file and then changed their mind about saving it.
Screenshot from 2026-08-07 18-23-23
Screenshot from 2026-08-07 18-23-46

### The file was automatically pulled into the input, as if the user had selected it. Previously, the input would have remained empty.
Screenshot from 2026-08-07 18-24-03

### We delete the existing value of the optional field using the button
Screenshot from 2026-08-07 18-24-11

### The optional file was successfully deleted.
Screenshot from 2026-08-07 18-24-23

## Other
* Legacy conditions in the `_handle_form_data` function are no longer needed

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import inspect
-import io
 import logging
 import warnings
 from collections.abc import Awaitable, Callable, Sequence
@@ -838,7 +837,7 @@ class Admin(BaseAdminView):
                 request, model_view.edit_template, context
             )
 
-        form_data = await self._handle_form_data(request, model)
+        form_data = await self._handle_form_data(request)
         form = Form(form_data)
         context["form"] = form
 
@@ -1037,30 +1036,12 @@ class Admin(BaseAdminView):
         return request.url_for("admin:create", identity=identity)
 
     @staticmethod
-    async def _handle_form_data(request: Request, obj: Any = None) -> FormData:
-        """
-        Handle form data and modify in case of UploadFile.
-        This is needed since in edit page
-        there's no way to show current file of object.
-        """
-
+    async def _handle_form_data(request: Request) -> FormData:
         form = await request.form()
         form_data: list[tuple[str, str | UploadFile]] = []
         for key, value in form.multi_items():
-            if not isinstance(value, UploadFile):
-                form_data.append((key, value))
-                continue
+            form_data.append((key, value))
 
-            should_clear = form.get(key + "_checkbox")
-            empty_upload = len(await value.read(1)) != 1
-            await value.seek(0)
-            if should_clear:
-                form_data.append((key, UploadFile(io.BytesIO(b""))))
-            elif empty_upload and obj and getattr(obj, key):
-                f = getattr(obj, key)  # In case of update, imitate UploadFile
-                form_data.append((key, UploadFile(filename=f.name, file=f.open())))
-            else:
-                form_data.append((key, value))
         return FormData(form_data)
 
     @staticmethod
--- a/sqladmin/statics/js/main.js
+++ b/sqladmin/statics/js/main.js
@@ -699,3 +699,51 @@ function formatImportTemplate(template, values) {
     if (name !== null) rememberCategory(name, false);
   });
 })();
+
+// Event for the clear button of the optional file input
+$(document).on('click', '[data-role="clear-file-input"]', function () {
+    const fieldId = $(this).data('field-id');
+    if (!fieldId) return;
+
+    const $fileInput = $(fieldId);
+    if (!$fileInput.length || !$fileInput.is('input[type="file"]')) return;
+
+    $fileInput.wrap('<form>').closest('form').get(0).reset();
+    $fileInput.unwrap();
+});
+
+// Function for loading a file from base64 to a value in the file input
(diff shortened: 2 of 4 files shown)


sqladmin/widgets.py
```diff
@@ -131,27 +132,53 @@ class FileInputWidget(widgets.FileInput):
...
+        current_value = Markup("")
+
+        if isinstance(field.data, StorageFile):
+            filename = field.data.name
+            current_value = Markup("Currently: {value}").format(
+                value=escape(field.data.path)
             )
-        else:
-            checkbox = Markup()
+            base64_bytes = base64.b64encode(field.data.open().read()).decode("utf-8")
```
mmzeynalli, line 144: There is still no size guard: a 50 MB stored file becomes ~67 MB of base64 embedded in the edit page's HTML, loaded into the browser's DOM. What if we do capping at a reasonable size (e.g. 2–5 MB) and falling back to the "Currently: path" text-only display for larger files?

sqladmin/statics/js/main.js
```diff
@@ -699,3 +699,51 @@ function formatImportTemplate(template, values) {
...
+
+// Event for the clear button of the optional file input
+$(document).on('click', '[data-role="clear-file-input"]', function () {
+    const fieldId = $(this).data('field-id');
+    if (!fieldId) return;
+
+    const $fileInput = $(fieldId);
+    if (!$fileInput.length || !$fileInput.is('input[type="file"]')) return;
+
+    $fileInput.wrap('<form>').closest('form').get(0).reset();
```
mmzeynalli, line 711: This creates a nested <form> inside the existing page <form>, which is invalid HTML. It works in practice because .reset() is synchronous and .unwrap() immediately follows, but it is not good design.

```
fileInput.value = '';
// or
const dt = new DataTransfer();
fileInput.files = dt.files;
```

sqladmin/statics/js/main.js
```diff
@@ -699,3 +699,51 @@ function formatImportTemplate(template, values) {
...
+    $fileInput.unwrap();
+});
+
+// Function for loading a file from base64 to a value in the file input
+function handlePreloadedFile(base64Data, fieldId, filename) {
```
mmzeynalli, line 716: not async, awaited in line 748. Make this async

sqladmin/statics/js/main.js
```diff
@@ -699,3 +699,51 @@ function formatImportTemplate(template, values) {
...
+  const binaryString = window.atob(base64Data);
```
mmzeynalli, line 717: have this in try/catch

sqladmin/statics/js/main.js
```diff
@@ -699,3 +699,51 @@ function formatImportTemplate(template, values) {
...
+  const len = binaryString.length;
+  const bytes = new Uint8Array(len);
+  for (let i = 0; i < len; i++) {
+    bytes[i] = binaryString.charCodeAt(i);
+  }
+  const blob = new Blob([bytes], { type: 'application/octet-stream'});
```
mmzeynalli, line 723: The real MIME type is discarded. The backend probably doesn't care, but if any browser-side validation or accept= attribute filtering is used on the <input>, the wrong type will trip it.

sqladmin/application.py
```diff
@@ -1037,30 +1036,12 @@ def get_save_redirect_url(
...
-            if should_clear:
-                form_data.append((key, UploadFile(io.BytesIO(b""))))
-            elif empty_upload and obj and getattr(obj, key):
-                f = getattr(obj, key)  # In case of update, imitate UploadFile
-                form_data.append((key, UploadFile(filename=f.name, file=f.open())))
-            else:
-                form_data.append((key, value))
+            form_data.append((key, value))
+
         return FormData(form_data)
```
mmzeynalli, line 1045: Instead op looping through all keys and copying:

```py
async def _handle_form_data(request: Request) -> FormData:
    form = await request.form()
    form_data: list[tuple[str, str | UploadFile]] = []
    for key, value in form.multi_items():
        form_data.append((key, value))
    return FormData(form_data)
```

mmzeynalli (changes requested):
What I noticed:

## 1

In `helpers.py`, `validate_servable_file_path` falls back to `Path.cwd().resolve()` when `allowed_roots` is empty:

`get_column_storage_roots()` returns `[]` for any plain String column (e.g. `Asset.url`). So if a record stores a local path like `../../config.py` in that column — bypassing `CDNURLField` via a direct API call or a raw DB write — `get_object_filepath` will:

See it's not an `http://` URL → skip the 400 guard
Call `validate_servable_file_path(path, [])`→ root becomes `cwd` → serves any file under the project root

Fix: when `allowed_roots` is empty and the value is not an HTTP URL, raise 403/400 rather than falling back to cwd.


## 2

fields.py — "FileField" appears on lines 31 and 38. The second entry is where "BooleanField" should be listed.

## 3

`value_is_filepath` does a blocking `os.path.isfile()` on the async event loop
Called once per row during list-page rendering — blocks the event loop for each stat call. Should be replaced with anyio.Path(...).is_file() or run via anyio.to_thread.run_sync.

## 4

Add clear file to non-optional fields as well. User might want to clear, and then upload.

mmzeynalli:
@maxim-f1 any updates?

PR #1125, opened 2026-08-08
Added debug_toolbar for sqladmin

### LINKED ISSUE #1053: Enable fastapi-debug-toolbar integration
### Discussed in https://github.com/smithyhq/sqladmin/discussions/820

<div type='discussions-op-text'>

<sup>Originally posted by **mariofix** September 17, 2024</sup>
I was wondering if any of you uses fastapi-debug-toolbar, I get the panel within /docs and /redoc or when i load a template but it doesn't load when using the Admin</div>

haykeminyan:
Closes #1053


docs/debug_toolbar.md
```diff
@@ -0,0 +1,79 @@
+```markdown
```
mmzeynalli: remove this line

docs/debug_toolbar.md
```diff
@@ -0,0 +1,79 @@
...
+
+class User(Base):
+    __tablename__ = "users"
+    id = Column(Integer, primary_key=True)
+    name = Column(String)
+
+Base.metadata.create_all(engine)
+
+app = FastAPI()
+admin = Admin(app, engine)
```
mmzeynalli: There is no documentation for debug_toolbar param

docs/cookbook/debug_toolbar.md
mmzeynalli, line 1: Add this file to mkdocs.yml as well

mmzeynalli: Also document 
```
# For Litestar
pip install debug-toolbar[litestar]

# For Starlette
pip install debug-toolbar[starlette]

# For FastAPI
pip install debug-toolbar[fastapi]
```
Saying that user can install any of these. And we need to test starlette mainly

sqladmin/application.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         logo_height: int = 64,
         favicon_url: str | None = None,
         templates_dir: str = "templates",
+        debug_toolbar: bool | Any = False,
```
mmzeynalli: Add this after audit_backend, like you did in others

mmzeynalli (changes requested):
Tests are missing

docs/debug_toolbar.md
mmzeynalli: and last 3 ticks

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
   "jinja2",
   "python-multipart",
   "sqlalchemy >= 2.0",
+  "litestar-debug-toolbar>=0.4.2",
```
mmzeynalli: ?

mkdocs.yml
```diff
@@ -39,6 +39,7 @@ nav:
       - Multiple databases: "cookbook/multiple_databases.md"
       - Using rich text editor: "cookbook/using_wysiwyg.md"
       - Audit logging: "cookbook/audit_logging.md"
+      - Debug toolbar: "docs/debug_toolbar.md"
```
mmzeynalli: Put it under cookbook

pyproject.toml
mmzeynalli: make it optional dependency like sqladmin[debug]

mmzeynalli (changes requested):
There is also this issue:

With only `litestar-debug-toolbar` installed, the toolbar page loads but its
stylesheet and script do not:

```
/admin/_debug_toolbar/                   -> 200
/admin/_debug_toolbar/static/toolbar.js  -> 500
/admin/_debug_toolbar/static/toolbar.css -> 500
```

This is an upstream packaging issue The Starlette asset
route imports its handler from the Litestar subpackage:

```python
# debug_toolbar/starlette/routes.py
def _get_toolbar_js() -> str:
    from debug_toolbar.litestar.routes.handlers import get_toolbar_js
    return get_toolbar_js()
```

and `debug_toolbar/litestar/__init__.py` imports `litestar` at module level, so
the import fails with `ModuleNotFoundError: No module named 'litestar'` on any
non-Litestar install. The handler itself just returns a string and needs nothing
from Litestar.

Until that is fixed upstream, we keep this PR open. We can even contribute to that project




ALSO

Image

This UI is messy, we need to integrate that fully.




Furthermore:
Why one would need debug toolbar in admin? You would usually need it in docs (i.e. swagger) no? To test optimization of your endpoints. One rarely uses BaseViews, and do not need separate debug toolbar just to see 1-2 page worth stuff. Im considering if we need this feature at all

mmzeynalli (approved):

PR #1126, opened 2026-08-19
fix: handle translated save button labels in create/edit redirects

x23n5902y:
In non-English locales, submitting an edit or create form in SQLAdmin redirects the user to the empty create page instead of the expected list page. Additionally, the "Save as new" button on the edit page updates the current record instead of creating a copy. Saving itself always succeeds — the record is written to the database — only the post-save navigation (and the copy semantics) are wrong.

This only reproduces when the admin UI is rendered in a non-English locale (I18nConfig(default_locale="ru"), or when the locale is negotiated from the Accept-Language header / language cookie / ?lang= query parameter). With the default "en" locale everything works as expected.

Add a small private helper on Admin that matches the submitted button value against both the English literal and the active-locale translation:

```python
@staticmethod
def _save_button_clicked(form: FormData, label: str) -> bool:
    """Return whether the submit button ``label`` was pressed.

    The create/edit templates render the submit buttons via gettext, so in
    non-English locales the submitted ``save`` value is the translated
    label. Compare against both the English literal and the active-locale
    translation.
    """
    save = form.get("save")
    return save in (label, gettext(label))
```

gettext resolves against the request locale set by LocaleMiddleware — the same locale that was used to render the button in the first place — so the comparison always matches what the user actually clicked. The English literal is kept in the comparison, so the behavior for the default locale (and for apps without i18n) is unchanged.

### DIFF
--- a/sqladmin/application.py
+++ b/sqladmin/application.py
@@ -852,7 +852,9 @@ class Admin(BaseAdminView):
 
         form_data_dict = self._denormalize_wtform_data(form.data, model)
         try:
-            if model_view.save_as and form_data.get("save") == "Save as new":
+            if model_view.save_as and self._save_button_clicked(
+                form_data, "Save as new"
+            ):
                 obj = await model_view.insert_model(request, form_data_dict)
             else:
                 obj = await model_view.update_model(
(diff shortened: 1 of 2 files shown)


sqladmin/application.py
```diff
@@ -1023,19 +1025,32 @@ def get_save_redirect_url(
...
             url = URL(str(request.url_for("admin:list", identity=identity)))
             if request.url.query:
                 url = url.replace(query=request.url.query)
             return url
 
-        if form.get("save") == "Save and continue editing" or (
-            form.get("save") == "Save as new" and model_view.save_as_continue
+        if Admin._save_button_clicked(form, "Save and continue editing") or (
+            Admin._save_button_clicked(form, "Save as new")
+            and model_view.save_as_continue
```
mmzeynalli, line 1036: instead of this, make `_save_button_clicked` a global function (on top of the file) and use that.

tests/test_application.py
```diff
@@ -14,6 +14,7 @@
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 from sqladmin import Admin, ModelView
+from sqladmin.i18n import DEFAULT_LOCALE, set_locale
 from tests.common import sync_engine as engine
```
mmzeynalli, line 19: Add this

```py
def _translated(locale: str, label: str) -> str:
    """The label as a user in ``locale`` would see it on the button."""

    set_locale(locale)
    try:
        return gettext(label)
    finally:
        set_locale(DEFAULT_LOCALE)
```

tests/test_application.py
```diff
@@ -228,6 +229,49 @@ class UserAdmin(ModelView, model=User):
...
+    response = client.post(
+        "/user", data={"save": "Сохранить и продолжить редактирование"}
+    )
+    assert response.text == "http://testserver/admin/user/edit/1"
+
+    response = client.post("/user", data={"save": "Сохранить как новое"})
+    assert response.text == "http://testserver/admin/user/edit/1"
+
+    response = client.post("/user", data={"save": "Сохранить и добавить ещё"})
+    assert response.text == "http://testserver/admin/user/create"
```
mmzeynalli, line 272: And instead of hard coding russian equivalents:

```py
    def post(label: str) -> str:
        return client.post("/user", data={"save": _translated("ru", label)}).text

    assert post("Save") == "http://testserver/admin/user/list"
    assert post("Save and continue editing") == "http://testserver/admin/user/edit/1"
    assert post("Save as new") == "http://testserver/admin/user/edit/1"
    assert post("Save and add another") == "http://testserver/admin/user/create"
```

mmzeynalli: Also, just in case add test for "Save as new"

mmzeynalli (changes requested):
Nice finding! Some little buggy fixes needed, though...

PR #1137, opened 2026-09-03
fix: validate pagination bounds before constructing LIMIT/OFFSET

### LINKED ISSUE #1136: Unhandled OverflowError on oversized page parameter in list view
## Description

`GET /admin/<identity>/list?page=99999999999999999999` raises an unhandled `OverflowError` (Python int too large to convert to SQLite INTEGER) and returns a 500. The page value is clamped only after the SQL is built.

## Steps to Reproduce

1. Register a `ModelView` with a few rows (SQLite is enough).
2. Request `/admin/<identity>/list?page=99999999999999999999`.

## Actual Behavior

- `page=99999999999999999999` returns 500 (unhandled OverflowError)
- `page=-1` returns 200
- `page=0` returns 200
- `pageSize=-5` returns 200
- `pageSize=99999999999999999999` returns 200
- `page=abc` returns 400 (existing validation)

## Expected Behavior

The page should be clamped to the last valid page before the LIMIT/OFFSET is built, matching the existing redirect behavior for oversized pages. Or the request should be rejected with 400.

## Environment

- sqladmin 0.31.1 (commit 7161c8a, HEAD of main)
- Python 3.12, SQLite

## Root Cause

`validate_page_number` (sqladmin/models.py:1022-1032) accepts any int. `ModelView.list` (sqladmin/models.py:1040-1092) computes the offset at line 1082 (`stmt.limit(page_size).offset((page - 1) * page_size)`) before any clamping. `Pagination.__post_init__` (sqladmin/pagination.py:49-51) clamps `page` only after the SQL has already been built.

Vansh-Sharma27:
## Summary

`GET /admin/<identity>/list?page=99999999999999999999` raises an unhandled OverflowError (Python int too large to convert to SQLite INTEGER) and returns 500. The page value is clamped only after the SQL is built, so oversized values reach the database driver.

This PR clamps the page to the last valid page before LIMIT/OFFSET is constructed. Out-of-range page values redirect to the closest valid page instead of reaching the database, and non-positive `pageSize` values return 400.

## Changes

- `ModelView.list` constructs `Pagination` right after the count, before `LIMIT/OFFSET` is built. `Pagination.__post_init__` is the single source of page clamping, so the max-page calculation is no longer duplicated between models.py and pagination.py.
- `validate_page_number` only rejects non-integer values now. Requests below page 1 go through the same redirect path as oversized pages, so both out-of-range bounds behave the same way.
- `pageSize` values below 1 return 400. `Pagination` also raises `ValueError` when constructed with a non-positive `page_size`.
- Oversized positive `pageSize` values keep the existing `page_size_options` cap.

## Behavior

Before (0.31.1):

- `page=99999999999999999999` returns 500
- `page=-1` returns 200
- `page=0` returns 200
- `pageSize=-5` returns 200
- `pageSize=0` returns 200
- `pageSize=99999999999999999999` returns 200
- `page=abc` returns 400

After:

- `page=99999999999999999999` redirects (302) to the last valid page
- `page=-1` redirects (302) to page 1
- `page=0` redirects (302) to page 1
- `pageSize=-5` returns 400
- `pageSize=0` returns 400
- `pageSize=99999999999999999999` returns 200 (capped by `page_size_options`)
- `page=abc` returns 400
- `pageSize=abc` returns 400

This matches the intent of the earlier pagination-validation change and the later maintainer-approved clamp-and-redirect behavior. [#752](https://github.com/aminalaee/sqladmin/pull/752) [#814](https://github.com/smithyhq/sqladmin/pull/814)

## Tests

- Parametrized rejection tests for `page=aaaa`, `pageSize=aaaa`, `pageSize=0` and `pageSize=-5`
- Parametrized redirect tests for `page=-1`, `page=0` and the 20-digit page value, asserting 302 to page 1, page 1 and the last valid page respectively
- Pagination clamp tests: empty dataset, exact page-size multiple, partial final page, lower-bound clamping
- `Pagination` rejects a non-positive `page_size` with `ValueError`

Format, lint, test, build and docs build pass locally. 608 passed, 4 skipped. Coverage 98%.

Closes [#1136](https://github.com/smithyhq/sqladmin/issues/1136)


sqladmin/pagination.py
```diff
@@ -48,7 +48,8 @@ def next_page(self) -> PageControl:
 
     def __post_init__(self) -> None:
         # Clamp page
-        self.page = min(self.page, max(1, self.count // self.page_size + 1))
+        max_page = max(1, (self.count + self.page_size - 1) // self.page_size)
+        self.page = min(max(self.page, 1), max_page)
```
mmzeynalli, line 52: So we get smth like this:

page=0        -> 400 error page
page=10000    -> 302 -> ?page=3

I think we should have one logic for both out-of-ranges. Either we redirect to closest possible or return 400, you choose.

sqladmin/models.py
```diff
@@ -1079,6 +1087,9 @@ async def list(self, request: Request) -> Pagination:
             request, select(func.count()).select_from(stmt.subquery())
         )
 
+        max_page = max(1, (count + page_size - 1) // page_size)
+        page = min(page, max_page)
```
mmzeynalli, line 1091: This part is repeated in pagination.py. Should it?

mmzeynalli (changes requested):
Nice catch and nice fix! Some little fixes. Also, I think we should verify page_size to be more than 0 as well in this PR.

sqladmin/pagination.py
Vansh-Sharma27: Agreed. I went with redirecting to the closest valid page, since that preserves the existing behavior for oversized pages. Requests below page 1 now redirect to page 1, and requests above the maximum still redirect to the last page. The redirect condition now checks for any difference instead of only going over, and both bounds have tests.

sqladmin/models.py
Vansh-Sharma27: Good point. I removed the duplicate calculation from `ModelView.list`. `Pagination` is now constructed before `LIMIT/OFFSET` is built, so `Pagination.__post_init__` is the single source of page normalization. The query uses the normalized `pagination.page`, and the rows are assigned to the object after fetching.

mmzeynalli (approved):
Added minor fix, would have been harder to explain, than do it) Anyways, good job!

